PR #232, opened 2014-08-28
This fixes a problem for Rails Admin show and edit actions (and posibly for delete).

jvortmann:
This is done by turning the primary keys attributes from a string to an array by splitting on '/' (used on Rails Admin routes).


cfis:
So is this still outstanding or can I close?

jvortmann:
@cfis I think this can me merged. Without it, the Rails Admin routes would break.

cfis:
Don't know if this is still an issue but I merged this into master (finally!).

Thanks for the help!

Issue #257, opened 2014-11-11
CPK generate invalid SQL syntax with optimistic lock


kirikak2:
It looks like composite primary keys generate invalid SQL Syntax with the optimistic lock.

```
> book.save
   (0.2ms)  begin transaction
   (0.5ms)  UPDATE "books" SET "content" = 'test', "updated_at" = '2014-11-11 14:13:33.437735', "lock_version" = 1 WHERE ("books"."cpk1,cpk2" = '1', '2' AND "books"."lock_version" = 0)
SQLite3::SQLException: near ",": syntax error: UPDATE "books" SET "content" = 'test', "updated_at" = '2014-11-11 14:13:33.437735',
', '2' AND "books"."lock_version" = 0)
SQLite3::SQLException: near ",": syntax error: UPDATE "books" SET "content" = 'test', "updated_at" = '2014-11-11 14:13:33.437735', "lock_version" = 1 WHERE ("books"."cpk1,cpk2" = '1', '2' AND "bo
ks"."lock_version" = 0)
```

Books table has optimistic lock column(lock_version)
Is CPK not support optimistic lock?

Thanks.

kirikak2:
This problem is solved. Thank you.

PR #258, opened 2014-11-11
Support optimistic lock.

kirikak2:
#257


cfis:
Sorry for not getting to this before - and thanks for writing up.  Anyway you can use the existing fixtures though for the test cases?  And have you moved to Rails 4.2, should we apply it to that branch?

kirikak2:
No problem. Thank you for your response.
This test case needs lock_version column. However existing fixtures have no lock_version column. Can I add lock_version column to any tables?

This problem occurred in ar_4.1.x branch. (maybe ar_4.2.x branch too).
I want apply this patch to ar_4.1.x and others. What should I do?

cfis:
Yes, feel free to add lock-version column to whichever table(s) you want.  I'm ok with modifying the existing tables, adding/removing test data, etc.  I would just rather avoid adding more tables if its possible.  Of course if its not, then that is ok.

kirikak2:
Thank you. I fixed this, and created new PR #272.
I will check to apply other branches.

cfis:
Ok, great, will close this PR then.

Issue #262, opened 2014-12-05
Rails 4.2.0.rc1


mmahalwy:
I keep getting this error:

```
NoMethodError: undefined method `ensure_initialized' for #<ActiveRecord::AttributeSet:0x007fe44c2b52b8>
/Users/mmahalwy/.rvm/gems/ruby-2.1.2/bundler/gems/composite_primary_keys-0a8061943732/lib/composite_primary_keys/core.rb:6:in `block in init_internals'
/Users/mmahalwy/.rvm/gems/ruby-2.1.2/bundler/gems/composite_primary_keys-0a8061943732/lib/composite_primary_keys/core.rb:6:in `each'
/Users/mmahalwy/.rvm/gems/ruby-2.1.2/bundler/gems/composite_primary_keys-0a8061943732/lib/composite_primary_keys/core.rb:6:in `init_internals'
/Users/mmahalwy/.rvm/gems/ruby-2.1.2/gems/activerecord-4.2.0.rc1/lib/active_record/core.rb:300:in `init_with'
/Users/mmahalwy/.rvm/gems/ruby-2.1.2/gems/activerecord-4.2.0.rc1/lib/active_record/persistence.rb:69:in `instantiate'
/Users/mmahalwy/.rvm/gems/ruby-2.1.2/gems/activerecord-4.2.0.rc1/lib/active_record/querying.rb:50:in `block (2 levels) in find_by_sql'
/Users/mmahalwy/.rvm/gems/ruby-2.1.2/gems/activerecord-4.2.0.rc1/lib/active_record/result.rb:51:in `block in each'
/Users/mmahalwy/.rvm/gems/ruby-2.1.2/gems/activerecord-4.2.0.rc1/lib/active_record/result.rb:51:in `each'
/Users/mmahalwy/.rvm/gems/ruby-2.1.2/gems/activerecord-4.2.0.rc1/lib/active_record/result.rb:51:in `each'
/Users/mmahalwy/.rvm/gems/ruby-2.1.2/gems/activerecord-4.2.0.rc1/lib/active_record/querying.rb:50:in `map'
/Users/mmahalwy/.rvm/gems/ruby-2.1.2/gems/activerecord-4.2.0.rc1/lib/active_record/querying.rb:50:in `block in find_by_sql'
[...]

PR #263, opened 2014-12-06
Fixes the no method error for ensure_initialized

mmahalwy:
`ensure_initialized` method exists up to 4.2.0.beta4, but removed in 4.2.0.rc1, see: https://www.omniref.com/ruby/gems/activerecord/4.2.0.rc1/files/lib/active_record/attribute_set.rb#tab=Q---A

Fixed code in the same format as previous versions not using `ensure_initialized`


cfis:
I think this was fixed in #262.

PR #268, opened 2015-01-02
README change to convey finding available versions

aaronbartell:
The README says to specific a specific version in the Gemfile but then lists "6.x" and doesn't spell out what versions are available in the 6.x area.  I added a small blurb on how to display all available versions.


cfis:
Thanks!

PR #272, opened 2015-01-11
Support optimistic lock for ar_4.1.x

kirikak2:
please see #258.


cfis:
Thanks for the patch.  Would be great to have on AR 4.2 branch too.

PR #275, opened 2015-01-13
Support optimistic lock for ar 4.0.x

kirikak2:
Support optimistic lock for ActiveRecord 4.0.x
Original optimistic lock was overwritten by persistance.rb in CPK.


cfis:
Thanks!

Issue #224, opened 2014-07-15
Composite Keys Not Saving with SQL Server 2014


gbenedict:
I have a table with composite key made up of three fields.  When I attempt to save the record, it is failing.

```
2.1.2 :022 > user_role = UserRole.new(:VLSUserId => 1000, PartnerId: 1001, RoleId: 1002)
 => #<UserRole VLSUserId: 1000, RoleId: 1002, PartnerId: 1001> 

2.1.2 :023 > user_role.save
TinyTds::Error: Invalid column name 'RoleId'.: EXEC sp_executesql N'INSERT INTO [VLSUserRole] ([PartnerId], [RoleId], [VLSUserId]) OUTPUT inserted.VLSUserId,RoleId,PartnerId VALUES (@0, @1, @2)', N'@0 int, @1 int, @2 int', @0 = 1001, @1 = 1002, @2 = 1000
```

I believe the issue stems from [line 301 in the SQL Server adapter](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/blob/eaced2841bd91301fecec1f7f0b7e2ab5d1b4327/lib/active_record/connection_adapters/sqlserver/database_statements.rb#L301):

```
        def sql_for_insert(sql, pk, id_value, sequence_name, binds)
          sql =
            if pk
              sql.insert(sql.index(/ (DEFAULT )?VALUES/), " OUTPUT inserted.#{pk}")
            else
              "#{sql}; SELECT CAST(SCOPE_IDENTITY() AS bigint) AS Ident"
            end
          super
        end
```

The `OUTPUT` portion is specified incorrectly:

```
EXEC sp_executesql N'INSERT INTO [VLSUserRole] (PartnerId, RoleId, VLSUserId) OUTPUT inserted.VLSUserId,RoleId,PartnerId VALUES (@0, @1, @2)', N'@0 int, @1 int, @2 int', @0 = 1000, @1 = 1002, @2 = 1001
```

It should be:

```
EXEC sp_executesql N'INSERT INTO [VLSUserRole] (PartnerId, RoleId, VLSUserId) OUTPUT inserted.VLSUserId, inserted.RoleId, inserted.PartnerId VALUES (@0, @1, @2)', N'@0 int, @1 int, @2 int', @0 = 1000, @1 = 1002, @2 = 1001
```

When this change is made the insert works.

For reference, here is the relevant part of my Gemfile:

```
gem 'rails', '4.1.4'
[...]

PR #281, opened 2015-01-31
Monkey patch sql_for_insert for sqlserver adapter (issue #224)

jmozmoz:
(Not sure, why there is a problem with line end characters.)


cfis:
Yeah, looks good minus all the noise.  Do you mind trying to fix that.  Otherwise I guess I will apply by  hand.

jmozmoz:
Line ends look ok now.

cfis:
Thanks!

PR #282, opened 2015-02-03
Update README.rdoc

Davidslv:
I just wanted to update the README as I just experienced a conflict in my Gemfile when I tried to use version `7.0.0`

Thank you,
David Silva


cfis:
Thanks!

PR #286, opened 2015-02-08
Fix sqlserver adapter 4.2

jmozmoz:
This updates the pull request #281 (issue #224) to the current master (activerecord 4.2). With it, most of the test cases (see also new entry in databases.example.yml) pass for SQL Server 2012 Express

(I am still not sure, that issue #224 should be fixed in composite_primary_key and not in activerecord-sqlserver-adapter.)


cfis:
Thanks!

PR #289, opened 2015-02-10
Windows sqlserver test setup

jmozmoz:
Make it easier on Windows to run unit tests for sqlserver (still 4 fails left).


cfis:
Thanks!

PR #298, opened 2015-05-08
Case insensitive key compares in the preloader.

CyborgMaster:
It so happened that some of my data has differing cases for some of my keys.  I know this would be considered bad practice, but it is actually data synced from a client's database which is managed by a program that is not affiliated with my company in any way, so we have no control.  

It's for animals.  There is a `species` table with a string primary key, and they store it in all caps (Ex: CANINE, FELINE, etc).  It shows up in the `animals` table in all lower case (Ex: canine, feline, etc).  During a join between those two tables MySql is perfectly happy because it does all its compares case insensitive (as do most databases by default).  However, when preloading relations, CPK crashes.  

This isn't something that ActiveRecord handles correctly either, so I thought about submitting this change there, but Rails expects all of their primary keys to be integers, so they shouldn't have this problem.  However, composite keys are often necessary because you don't have a surrogate, auto-incrementing primary key.  In our case we have a composite key with a client identifier (to partition the data based on the client we synced it from), along with the string based primary key from their database.  So CPK has to deal with string keys quite frequently, while ActiveRecord never expects to.  

This patch simply downcases all keys in the preloader to get a case insensitive compare, which matches MYSQL database queries.  I wouldn't mind trying to figure out which database adapter is being used, to see if we should do case-sensitive or case-insensitive matches, but I don't know how to get that info.  If someone else does and wants to tell me how or add it themselves, that would be great.

### DIFF
--- a/lib/composite_primary_keys/associations/preloader/association.rb
+++ b/lib/composite_primary_keys/associations/preloader/association.rb
@@ -64,7 +64,7 @@ module ActiveRecord
             # }
             @preloaded_records.map { |record|
               key = Array(association_key_name).map do |key_name|
-                record[key_name]
+                record[key_name].to_s.downcase
               end.join(CompositePrimaryKeys::ID_SEP)
 
               [record, key]
@@ -77,7 +77,7 @@ module ActiveRecord
                                    # CPK
                                    # owner[owner_key_name].to_s
                                    Array(owner_key_name).map do |key_name|
-                                     owner[key_name]
+                                     owner[key_name].to_s.downcase
                                    end.join(CompositePrimaryKeys::ID_SEP)
                                  end
                                else
@@ -85,7 +85,7 @@ module ActiveRecord
                                    # CPK
                                    # owner[owner_key_name]
                                    Array(owner_key_name).map do |key_name|
-                                     owner[key_name]
+                                     owner[key_name].to_s.downcase
                                    end.join(CompositePrimaryKeys::ID_SEP)
                                  end
                                end


cfis:
Not following how the patch fixes the issue.  Doesn't the Array#join method already convert everything to a string?  So isn't the patch redundant?

CyborgMaster:
Yes, `Array#join` does convert everything to strings, but the problem isn't the string conversion, it is the down-casing.  MySQL does case insensitive compares on it's primary keys so we need to do so inside CPK as well.  You will notice that not only is `to_s` called, but also `downcase`.  This allows the case insensitive compares.

CyborgMaster:
Any opinion on the downcasing?

CyborgMaster:
bump.... 

My fork is getting a little out of date.  I would love to get this merged and abandon my fork.  Maybe I should just rebase my fork on master.

cfis:
Ah sorry to let this hang.  So I can't say I love this idea since its mysql specific and also seems like something the application using mysql should know about and can do itself.  Or like you said above, to do it at the database adapter level.  CPK does override a couple of the database adapters, see this folder:

https://github.com/composite-primary-keys/composite_primary_keys/tree/master/lib/composite_primary_keys/connection_adapters

Could the patch go in there?

CyborgMaster:
So I'm not sure if it's exactly MySQL specific.  Any database that does case insensitive compares would do it this way, and I'm pretty sure most databases can be set up that way, even if it isn't their default (it is the default in MySQL).  

As far as the application handling it, that sounds like a good solution, except for the fact that I don't know how it would be possible.  Because this is in the preloader, it happens between loading one set of records for an association, and another.  I don't know how the application could hook in at that point, they normally can't get ahold of the records until after the preloader is finished.  

I would be all for putting the patch in the database adapter, although I'm not sure how I would do that, seeing as how the important code is in the preloader.  Can you think of some way I could modify the preloader's behavior from within the adapter? 

Due to the fact that most (all?) databases can be set up for case insensitive compares, maybe this would be a good candidate for a global config.  That way people can set it up to match their database configuration.  Would that be possible?

CyborgMaster:
@cfis any further thoughts on my responses?  Does a cpk global config sound like a good option here?

cfis:
I'd still rather keep this mysql specific, so would still rather the database adapter approach.

CyborgMaster:
I am happy to put this in the mysql adapter.  I'm not sure exactly how I would do that, seeing as how the important code is in the preloader. Can you think of some way I could modify the preloader's behavior from within the adapter?

CyborgMaster:
@cfis any thoughts on my above question?

CyborgMaster:
I'm rebasing these changes onto the latest changes from master.  I'm still open to doing a MySQL specific change, but I don't know how to include it in the adapter seeing as how the changed code is the in the preloader.  Maybe this would be a better global config so anyone who has set up their database for case insensitive compares can use it (not just MySQL).  

Any thoughts on how I could do it in the adapter or a global config?

CyborgMaster:
I'm really sorry that I keep bumping this thread, but I don't know how to proceed without additional direction.  Help?

cfis:
Sorry, finally had some time to look at this again.  I notice that the abstract adapter has a method called can_perform_case_insensitive_comparison_for?

Can you update the patch to use that method to check if its is a string column, is case insensitive, and if yes then use lower?

CyborgMaster:
AWESOME!  That sounds perfect.  I'm sorry that I didn't know about that method myself, but I'm not very familiar with the internals.  I'll make that update and ping you when its done.  It will take me a bit because I'm heads down on other projects right now.  

Thanks for the tip on that method!

cfis:
Any progress on this?

Issue #306, opened 2015-06-25
7.0.13 / ActiveRecord 4.1, composite primary key where one part is nil not supported


rud:
Hi,

So, during the upgrade of an app I find that a key like `[nil, 132]` no longer works with a basic `.find()`, it always finds 0 records rather than the record where `first_part IS NULL AND second_part = 132`. This worked fine in version 6.0.8 (4.0.x) and 5.0.14 (3.2.x).

In effect I've got a schema like this:
```
create table mittens (
  left_id int(11),
  right_id int(11) not null
  name varchar(50) not null
  primary key (left_id, right_id)
);
```

How to test for this - should I spawn another test-table and add a fixture for it? I know it's a corner-case, but I don't mind trying to add support for it.

rud:
Ended up introducing a new model `Mitten` and a fixture with a nil as part of the primary key. I've run it on Travis CI, with this result: https://travis-ci.org/rud/composite_primary_keys/builds/68321336

PR #307, opened 2015-06-25
Bugfix/ar 4.1 allow nil in parts of composite primary key

rud:
This should be a test-case demonstrating issue #306, but I've had some trouble getting everything stable. At least the new test-case in `test_find.rb` correctly fails to find the intended record.

Any feedback at all is most welcome.


cfis:
Hi Rud - Not sure this is a good idea.  Postgresql does not support null values in composite keys (nor for that matter does Oracle).  Thus this would be adding a feature, with its extra complexity, that would only work on some databases (I assume mysql basically).

rud:
I realize this is potentially a controversial feature, and the database design is beautifully legacy. The origins of this feature-branch is the code actually worked as intended in rails 3.2 and 4.0 with post Oracle and sqlite (with ActiveRecord constructing the correct queries). Since this is a functionality used in production with a previous version of Rails, upgrading said Rails application is going to be somewhat tricky as `composite_primary_keys` has inadvertently lost this edge-case functionality, since nothing tested the behaviour in previous Rails-versions.

If that makes sense at all: nothing tested this, a project relies on the old behaviour, the behaviour is currently gone, but could fairly easily be reinstated, and there's even test-cases documenting how it should be working.

Just thought I'd clarify.

rud:
Btw, #309 is the base-line testing on Rails 3.2 where this use-case worked as intended.

deepredsky:
I haven't been able to upgrade to 4.1 due to this. It was working perfectly with nil values in composite keys in my Oracle. Would be great if this support was added back in, perhaps on a per db adapter basis to avoid issues with other adapters

deepredsky:
Looks like the issue here is with using bind values. The query generated with nil in primary keys is something like this

```
SELECT  "mittens".* FROM "mittens"  WHERE ("mittens"."left_id" = $1 AND "mittens"."right_id" = $2) LIMIT 1  [["left_id", nil], ["right_id", 3]]
```

If instead we use arel predicates without using  `bind_values`, it will correct nil to us 'IS NULL' and the problem is fixed.

```
SELECT  "mittens".* FROM "mittens"  WHERE ("mittens"."left_id" IS NULL AND "mittens"."right_id" = 3) LIMIT 1
```

rud:
Sounds exact right @deepredsky. In hindsight, cursing for not catching it myself. 

You feel up for fixing it? I'm not active on a project that depends on this currently, so no commits or tests will flow from me, but please run with it if you are able :)

deepredsky:
@rud  I did a quick path removing the bind_values. It seems to work for my and my project. I even rebased the path on top of your test setup and it all seems to be working. but dunno what bug i might have introduced.

https://github.com/composite-primary-keys/composite_primary_keys/compare/ar_4.1.x...deepredsky:bugfix/allow-nil-composite-primary-keys?expand=1

Perhaps, you can check this and verify the fix.

mgeyer:
Hey guys, any news whether this will be implemented? It's preventing us from properly destroying records.

cfis:
Hmm, this ticket just has test changes (plus other stuff), not actual code changes.  I didn't like the original idea of allowing nulls, but it sounds like a different approach has been figured out.  But there is not push request with those changes.  So...I'm going to close this and if someone wants to create a pull request with what is needed that would be great.

PR #309, opened 2015-06-26
Chore/ar 3.2.x: baseline testing: allow nil in parts of composite primary key

rud:
I'm on a bit of a quest here with #306, so I thought I'd establish a proper test of the functionality in the AR 3.2 supporting version of the gem.

As a bonus I include a working Travis CI setup, ready for enabling, should you wish to do so. I can also make a PR without anything but the `mittens` stuff. 

What do you all think?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ db
 *.swp
 
 test/connections/databases.yml
+Gemfile.lock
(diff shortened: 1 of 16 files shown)


rud:
Build test up at https://travis-ci.org/rud/composite_primary_keys/builds/68452561

cfis:
See #307 for discussion.

PR #310, opened 2015-06-26
master: AR 4.2.x green Travis CI

rud:
This took some prodding, but I finally succeeded in making an [all-green build on Travis CI](https://travis-ci.org/rud/composite_primary_keys/builds/68489607) against `sqlite3`, `postgresql`, and `mysql`.

I think this is cause for celebration, what do you all think? Time to add a Travis CI logo to the README so things stay green?

PR #311, opened 2015-06-26
AR 4.2.x support compound keys with nil

rud:
This is based on #310 (establish green CI). It introduces a test to ensure `nil` is a valid value in a compound foreign key. For now this is just a failing test, as this has been broken since ActiveRecord 4.1 (worked fine in ActiveRecord 3.2.x and 4.0.x). 

Other related pull-requests and issues:
- #306 - original issue where I documented the changed behaviour
- #309 - prove it works in ActiveRecord 3.2
- #307 - a similar patch-set for ActiveRecord 4.1 with a failing test

Sorry for all the noise and issues and pull-requests with various bases, but I'd love for help to resolve this, if anybody has any idea where to start fixing the code. For now we have a failing test-case demonstrating the issue (45e4bb, which passes in gem-versions supporting ActiveRecord 3.2 and 4.0). 

I promise there will be cake :cake:

### DIFF
--- a/test/fixtures/db_definitions/db2-create-tables.sql
+++ b/test/fixtures/db_definitions/db2-create-tables.sql
@@ -124,3 +124,9 @@ create table products_restaurants (
   franchise_id integer not null,
   store_id integer not null
 );
+
+create table mittens (
+  left_id INTEGER,
+  right_id INTEGER NOT NULL,
+  name varchar(50) NOT NULL
+);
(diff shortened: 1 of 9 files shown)


cfis:
Hi Rud - thanks for all the work on this.  But see my comment on #307 - I don't think this is a good idea.

cfis:
Closing - discussion in #307 which came up with a different approach.

Issue #294, opened 2015-04-10
:verify_active_connections! removed


nextbyte:
:verify_active_connections! was removed from Rails 4 but it is still referenced in connection_specification_changes.rb 

      delegate :clear_reloadable_connections!,
        :clear_all_connections!,:verify_active_connections!, :to => :connection_handler

Which causes errors with Spawnling for example which uses this:

ActiveRecord::Base.verify_active_connections! if ActiveRecord::Base.respond_to?(:verify_active_connections!)

ActiveRecord::Base.respond_to?(:verify_active_connections!) will return true because :verify_active_connections! is added in connection_specification_changes.rb

cfis:
Ah, thanks.  Will remove.

PR #314, opened 2015-07-07
Do not delegate verify_active_connections! if method is undefined

ThePletch:
Addresses #294.


cfis:
Thanks for the patch.  Since each version of CPK is tied to a rails version, I don't think we need the conditional check.  Instead, we can just remove the code entirely like in [ar_4.1.x 18c8478] .

PR #315, opened 2015-07-09
Optimize find_some to hit database only once

rmello:
There is a huge database load when finding using a large array of CPKs due to the multiple DB queries. This PR makes a single query - approaching the active record way of finding multiple records

### DIFF
--- a/lib/composite_primary_keys/relation/finder_methods.rb
+++ b/lib/composite_primary_keys/relation/finder_methods.rb
@@ -130,7 +130,7 @@ module CompositePrimaryKeys
         # CPK
         # result = where(table[primary_key].in(ids)).to_a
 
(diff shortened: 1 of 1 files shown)


cfis:
Hmm, do all databases that AR includes support binding multiple columns in a where clause like you are doing?  Is the syntax the same across all databases?

http://stackoverflow.com/questions/6672665/doing-a-where-in-on-multiple-columns-in-postgresql

cfis:
Did a bit more research, this is not supported by sql server.  See https://msdn.microsoft.com/en-us/library/ms177682.aspx.

Seems like other DBs do support this though, so a bit of problem.  Could:

* Check what db is in use and act appropriately
* Use a subquery
* Use an in memory table created via VALUES (or maybe that is a postgresql only thing).

PR #321, opened 2015-08-23
Fall back on old OR style if too many predicates

### LINKED ISSUE #320: cpk_or_predicate stack level too deep
When I have code hit https://github.com/composite-primary-keys/composite_primary_keys/blob/master/lib/composite_primary_keys/composite_predicates.rb#L11-L18

It generates a tree, where all elements go down the right side:

```
OR
  \
    OR
      \
        OR
```

This can cause a stack overflow when trying to build the SQL. In our case we had 500 or clauses and we were already about 200 levels deep in the stack, so boom, overflow. I am going to start working on a PR but no promises =). 

When I replaced this method with the old code that was there before we stopped seeing stack overflows.

brgaulin:
This resolves #320 where if you have too many items in your OR table, you can have a stack overflow.

Open to debate on how to solve this issue. right now I have done a fall-back, but maybe we should always do that instead? Let me know what everyone thinks =)

### DIFF
--- a/composite_primary_keys.gemspec
+++ b/composite_primary_keys.gemspec
@@ -27,4 +27,9 @@ Gem::Specification.new do |s|
   # Dependencies
   s.required_ruby_version = '>= 1.9.3'
   s.add_dependency('activerecord', '~>4.1.7')
+
+  # Needed for dev / test
+  s.add_development_dependency('sqlite3')
+  s.add_development_dependency('pg')
+  s.add_development_dependency('mysql2')
 end
--- a/lib/composite_primary_keys/composite_predicates.rb
+++ b/lib/composite_primary_keys/composite_predicates.rb
@@ -4,17 +4,16 @@ module CompositePrimaryKeys
       if predicates.length == 1
         predicates.first
       else
-        Arel::Nodes::And.new(predicates)
+        ::Arel::Nodes::And.new(predicates)
       end
     end
 
-    def cpk_or_predicate(predicates)
-      or_predicate = predicates.map do |predicate|
-        ::Arel::Nodes::Grouping.new(predicate)
-      end.inject do |memo, node|
-        ::Arel::Nodes::Or.new(memo, node)
+    def cpk_or_predicate(predicates, table = nil)
+      if predicates.size > 100
+        cpk_or_predicate_manual(predicates, table)
+      else
+        cpk_or_predicate_arel(predicates)
       end
-      ::Arel::Nodes::Grouping.new(or_predicate)
     end
 
     def cpk_id_predicate(table, keys, values)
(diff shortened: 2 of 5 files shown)


brgaulin:
Looks like this branch won't pass travis ci =/
```
Gem::LoadError: Specified 'mysql2' for database adapter, but the gem is not loaded. Add `gem 'mysql2'` to your Gemfile (and ensure its version is at the minimum required by ActiveRecord).
```
But tests do pass locally

brgaulin:
This is only for the 4.1 branch, but it should probably be applied in the 4.2/master branches if it looks good. Really the issue we had was only at 360 items in the OR clause, which is quite a few but not totally unreasonable. Maybe earlier on active record / cpk should do a union table or something if doing so many. But this does it for now.

nddeluca:
:+1:

I run into this problem when using `includes(...)` or `preload(...)` with larger active record queries.  My workaround has been to use `eager_load(...)` which uses a join to load the associations, avoiding the `AND/OR` statements.  I also find that building a long query string with `AND` and `OR`, tends to take longer than loading associations through a join too.

Another fix that should also have some performance improvements, would be to use `IN` syntax with multiple columns. So, instead of:
```sql
SELECT * FROM test WHERE (key_1 = 3 AND key_2 = 4) OR (key_1 = 5 AND key_2 = 6) OR ...
```
The query could be written as:
```sql
SELECT * FROM test WHERE (key_1, key_2) IN ((3,4),(5,6),...)
```

I know this syntax works on PostgreSQL, but I'm not sure what other database/adapters support it.  Also, Arel doesn't support the syntax, so it would have to be done manually or added to Arel then patched here.  Either way, the it would be a bit more involved to implement.

brgaulin:
@nddeluca Thank you for the comment. Ya there are some other ways that are better to do this. But for now I had to make this patch just to get our code working again. In our case, one of the primary_keys is always the same value so we wouldn't even need to add that on to each record or do and-or, just do one big in for the second column

cfis:
Will this fix it? 9c09e4f?

brgaulin:
So long as it doesn't build a one sided tree :) before it would just keep
adding ors to the right. So a big enough tree would overflow
On Sun, Dec 13, 2015 at 10:46 PM Charlie Savage <notifications@github.com>
wrote:

> Will this fix it? 9c09e4f
> <https://github.com/composite-primary-keys/composite_primary_keys/commit/9c09e4f30f43ac69265e228f08a165ededa2f62a>
> ?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/composite-primary-keys/composite_primary_keys/pull/321#issuecomment-164334424>
> .
>
-- 
--
Brendan Gaulin

CyborgMaster:
Why was this closed?  This looks like a good solution to the stack overflow problem.

cfis:
Because I was hoping 9c09e4f30f43ac69265e228f08a165ededa2f62a fixed it.  Is that not the case?

CyborgMaster:
Based on your comment in the discussion at #320, it seems like it caused the OR tree to be half the depth, but still have an overflow problem.

nathansamson:
I have just opened #363 which should fix this problem for real (unless you have millions and millions of records).

Assuming the possible stack depth is 100, now the list of OR predicates can be 2**100 which is a large number.

CyborgMaster:
@nathansamson, I really like the approach you have taken here.  @cfis, what do you think?  I'd love to get his approach merged and into live use.

Issue #283, opened 2015-02-05
ConnectionNotEstablished error when db connection is on class other than ActiveRecord::Base


wconrad:
When a model's database connection is not via ActiveRecord::Base, and the model has a composite primary key, then the model raises "ActiveRecord::ConnectionNotEstablished".

# Versions

* composite_primary_key 7.0.13
* ruby-2.2.0
* active-record 4.1.9
* [oracle-enhanced-adapter HEAD c2d659](https://github.com/rsim/oracle-enhanced/commit/c2d65949799852b85a9479ec75dffeae33bd5bbf)

# What's this craziness about ActiveRecord::Base not having a connection?

I have a set of models that need to connect to a different database than the default Rails database.  These models are derived from an intermediate base class rather than directly from ActiveRecord::Base.

The intermediate base class:

```
module OpacsDb
  module Db
    # We establish the Opacs database connection on this class; that
    # connection is then shared by all the derived classes.
    class Base < ActiveRecord::Base
      self.abstract_class = true
    end
  end
end
```

A model:

```
module OpacsDb
  module Db
    class Org < Base
      self.table_name = "org"
      self.primary_key = [:or_customer, :or_rcc_code]
    end
  end
end
```

The connection is established with OpacsDb::Db::Base, e.g.

    OpacsDb::Db::Base.establish_connection(:opacs_test)

This works fine for any model which does not have a composite primary key.

Making a query using the the model having a composite primary key results in a stack trace:

```
> OpacsDb::Db::Org.count
ActiveRecord::ConnectionNotEstablished: ActiveRecord::ConnectionNotEstablished
/home/wayne/.rvm/gems/ruby-2.2.0/gems/activerecord-4.1.9/lib/active_record/connection_adapters/abstract/connection_pool.rb:541:in `retrieve_connection'
/home/wayne/lab/composite_primary_keys/upstream/lib/composite_primary_keys/connection_adapters/abstract/connection_specification_changes.rb:51:in `retrieve_connection'
/home/wayne/lab/composite_primary_keys/upstream/lib/composite_primary_keys/connection_adapters/abstract/connection_specification_changes.rb:33:in `connection'
/home/wayne/.rvm/gems/ruby-2.2.0/gems/arel-5.0.1.20140414130214/lib/arel/tree_manager.rb:22:in `visitor'
[...]

PR #322, opened 2015-08-28
ConnectionNotEstablished error when db connection is on class other than ActiveRecord::Base

dennisfeiock:
Similar to #283, this pull request fixes `ConnectionNotEstablished` errors when the database connection has been established on a class other than `ActiveRecord::Base`.  The explanation of the use case is the exact same as #283.  I ran into the bug when trying to run `Model.find_each` as opposed to `Model.count` as @wconrad describes.

The merge @wconrad provided was against ActiveRecord 4.1, whereas this one targets ActiveRecord 3.2.  One minor change is that `result.column_types.fetch(column_alias)` threw an undefined method error, so I reverted back to using `column_for(column_name)` as the code did originally.

Since the code is practically the same as the accepted merge of #284, I have not tested it beyond verifying that `Model.find_each` and `Model.count` works correctly.

### DIFF
--- a/lib/composite_primary_keys/relation/calculations.rb
+++ b/lib/composite_primary_keys/relation/calculations.rb
@@ -18,7 +18,6 @@ module CompositePrimaryKeys
         # Postgresql doesn't like ORDER BY when there are no GROUP BY
         relation = reorder(nil)
 
-        column_alias = column_name
         # CPK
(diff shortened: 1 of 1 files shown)


cfis:
Don't know if you still need this...but merged to the old back branch.

Thanks for submitting!

PR #325, opened 2015-09-14
Bug Fix: wrong database syntax when using inheritance 

CarlosRoque:
##### This is a bug fix for version 5.x 
When connecting to different databases using inheritance, the incorrect syntax is being generated when the databases are of different types.

example:
consider the following class NewBase
```ruby 
class NewBase < ActiveRecord::Base
      self.abstract_class = true
end
```
When first connecting ActiveRecord to mysql and then using NewBase to connect to sqlserver, 
sql updates and saves failed because the query generated was using mysql syntax instead of sqlserver. 

I didn't create a test because I wasn't sure how properly write a test for this bug

### DIFF
--- a/lib/composite_primary_keys/persistence.rb
+++ b/lib/composite_primary_keys/persistence.rb
@@ -20,7 +20,7 @@ module ActiveRecord
           where_hash[key.to_s] = self[key]
         end
 
-        # CPK      
+        # CPK
         #relation.bind_values = [[column, id]]
 
         relation = self.class.unscoped.where(where_hash)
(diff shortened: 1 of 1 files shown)


cfis:
Ok, version 5.x means you are using ActiveRecord 3.2, right?  In that case, please submit this against the ar_3.2.x branch.  Note though, that branch is quite unsupported at this time.  But I can merge the commit if you'd like.  Looks like the ar_4.0.x branch has the same issue, fyi.

CarlosRoque:
Thanks
I will push the changes to both branches

PR #330, opened 2015-11-02
ar_4.0.x Bug Fix: wrong database syntax when using inheritance

CarlosRoque:
This is a bug fix for ar_4.0.x

First reported here [https://github.com/composite-primary-keys/composite_primary_keys/pull/325]

When connecting to different databases using inheritance, the incorrect syntax is being generated when the databases are of different types.

example:
consider the following class NewBase
```ruby
class NewBase < ActiveRecord::Base
      self.abstract_class = true
end
```
When first connecting ActiveRecord to mysql and then using NewBase to connect to sqlserver,
sql updates and saves failed because the query generated was using mysql syntax instead of sqlserver.

### DIFF
--- a/lib/composite_primary_keys/persistence.rb
+++ b/lib/composite_primary_keys/persistence.rb
@@ -3,7 +3,7 @@ module CompositePrimaryKeys
     module Persistence
       def relation_for_destroy
         return super unless composite?
-        
+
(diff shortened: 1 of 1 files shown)


cfis:
Thanks!

PR #341, opened 2016-03-01
"uninitialized constant" when using a has_many through association with a polymorphic source

tye:
When you have a has_many :through association with a polymorphic source:

```ruby
class User < ActiveRecord::Base
  has_many :commenting_users, through: :comments, source: :commenter, source_type: 'User'
end

class Comment < ActiveRecord::Base
  belongs_to :commenter, polymorphic: true
end
```

Trying to set the list of associated objects will fail with a "Uninitialized constant" error:

```ruby
User.find(1).update_attributes commenting_users: [User.find(2), User.find(3)]
```

will result in an exception like:

```ruby
NameError: uninitialized constant Comment::Commenter
```

See the test case I added for a working example of this failure.

In the HasManyThroughAssociation#delete_records method we are looking up the source class on the source_reflection. It doesn't make sense to me to ever do this in the case source_reflection is polymorphic.

Rails forces you to declare a source_type on the association in this case. We should look up the klass on the originating association instead of on the source association.

### DIFF
--- a/lib/composite_primary_keys/associations/has_many_through_association.rb
+++ b/lib/composite_primary_keys/associations/has_many_through_association.rb
@@ -19,7 +19,8 @@ module ActiveRecord
         scope = through_association.scope
         # CPK
         # scope.where! construct_join_attributes(*records)
-        if source_reflection.klass.composite?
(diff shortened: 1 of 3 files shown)


cfis:
Thanks!

PR #342, opened 2016-03-02
fix association_cache method missing error

choihongil:
association_cache method deleted in rails5 that replace ```association_cache.key?``` to ```association_cached?```


cfis:
Thanks.  Note AR 5.0 is not supported yet...

PR #343, opened 2016-03-14
change autosave_association.rb so that association autosave can work

ypxing:
Without patching ```autosave_association.rb```, below case will fail with error
```TypeError: [:dorm_id, :room_id] is not a symbol nor a string```

```
    room = Room.new(dorm_id: 1000, room_id: 1001)
    room_assignment = RoomAssignment.new(student_id: 1000)
    room_assignment.room = room
    room_assignment.save
```

The change is to fix the issue by using ```record.read_attribute``` instead of ```record.send``` in ```autosave_association.rb```.


ypxing:
I mixed one PR with another WIP. Just close them first and will pull later.

PR #345, opened 2016-03-14
Fix failure case test_delete_all_with_joins and make CPK work for Arel::Attributes::Attribute

ypxing:
The change patch Arel so that ```Arel::Attributes::Attribute``` can be handled well when #name is one Array.

It can fix the failure case test_delete_all_with_joins. (SQLite cannot work for this case since multi-column IN clause like (column1, column2) IN (...) is not allowed in SQLite3).


cfis:
Awesome - glad that failure is fixed.

PR #346, opened 2016-03-16
WIP: Find in batches for mysql, postgresql, sqlite3

apurvis:
Could not think of better answer to `find_in_batches` in the case of composite primary keys but this convoluted recursive `CASE` statement actually does the trick.  

This works because because the PK is the order on the disk, and so you can use `offset(n).limit(1)` to find the PK every `n` records, and then use it to compare.

In the case of the `Capitol` example with a batch_size of 2, you get two queries:
```sql
SELECT `capitols`.* FROM `capitols` WHERE (CASE WHEN country = 'Canada' THEN city >= 'Ottawa' ELSE country > 'Canada' END ) AND (CASE WHEN country = 'France' THEN city <= 'Paris' ELSE country < 'France' END )
SELECT `capitols`.* FROM `capitols` WHERE (CASE WHEN country = 'Mexico' THEN city >= 'Mexico City' ELSE country > 'Mexico' END ) AND (CASE WHEN country = 'The Netherlands' THEN city <= 'Amsterdam' ELSE country < 'The Netherlands' END )
```

This appears to work with mysql, postgres, and sqlite3.

@cfis


cfis:
What's the issue you are trying to resolve here?  Find by batch doesn't with the CPK?

apurvis:
yeah, basically.  Not only does it not work, but it _seems_ to work and
then it just skips records.

cfis:
Ok, so you did this against the master branch which is meant for Rails 5. At least in Rails 5 the implementation is quite different than what you have here.  Looks like the place to look is the method:

def in_batches(of: 1000, start: nil, finish: nil, load: false)

So was this meant for Rails 4.2.* or 5.*

apurvis:
this was tested against rails 3/4, not rails 5.

cfis:
Thanks.  Manually merged in 3497be5b168f4e81ac0f66700e654429f4833359 to ar_4.2.x branch.

PR #347, opened 2016-03-16
Some fixes for Oracle and functionality to mix composite pkey tables with one column pkey tables

areman:


cfis:
Wow, that's a big change set.  Why are so many changes needed for Oracle support?

areman:
Dear cfis,
the fixes for oracle are not the main content of this pull request. 
In our project are using table with composite pkey and tables with single column pkey. And the tables are joined in our models, too. 
Example 
master ( cpk: country, name )
detail ( cpk: country, name, city )

As long as we do it in this ways everything is fine, but now we have other table with a single column join to the master or detail table.
attribute ( pk: id )
master_id
The belongs_to releation attribute.master_id = master.id is currently not possible.
And this is the main change of this pull request. 

Best Regards from Germany

cfis:
First, thanks for all the effort put into this. Would like to get this merged, but lots of questions:

* Is this request meant for Rails 5.x or Rails 4.2.x?  Since it was against master, I assume 5.0?  If so, please verify against the latest Rails 5 changes.

* Would it be possible to reuse more of the existing test tables?  I would like to remove tables, not add more.

* Can you reuse the existing test files if appropriate.  For example, there is already test_habtm.rb but this pull request then adds test_has_and_belongs_to_many_ids.rb. Seems better if they were merged together.

* It looks like this patch might merge in a few other things.  I see the change to autosave associations, and then checks to see if an attribute is named :id, etc.  Would be easier to deal with this one part a time.

Thanks - Charlie

areman:
Hello Charlie,
you're right. It's for 4.2.x the master is the wrong place. We'll delivery a 5.x branch, too :-) 
I'll check your other remarks and modify the request.

Thanks,
André

cfis:
Do you still want to try and get this merged?

cfis:
Happy to reopen if you can find the time to get back to this?

Issue #332, opened 2015-12-21
Any prevision of support for Active Record 5.0?


Kalunia:
Hi, about Rails 5 arriving in the coming days, when can we expect the gem to support the Active Record 5.0? There are already some codes, issues and discussions rolling?

I would know too if is much complicated to make the adjusts in order to adapt to the new Active Record. Maybe we could help with some contribution and make this gem available for the newest rails version!

Thanks for the attention!

cfis:
Sorry for the delay.  The master branch has changes to support Rails 5.  I haven't had a chance to verify them, so haven't released a gem yet.  

If you have time, that would be quite helpful to check them.

dreyks:
The master branch still requests 'activerecord', '~> 4.2.0' in gemspec

jmonteiro:
Confirmed, trying to install it with Rails 5.0.0.beta2 renders the following:

```
jmonteiro@pc ~/C/rails5project> bundle
Fetching git://github.com/composite-primary-keys/composite_primary_keys.git
Fetching gem metadata from https://rubygems.org/..........
Fetching version metadata from https://rubygems.org/...
Fetching dependency metadata from https://rubygems.org/..
Resolving dependencies...
Bundler could not find compatible versions for gem "activerecord":
  In snapshot (Gemfile.lock):
    activerecord (= 5.0.0.beta2)

  In Gemfile:
    composite_primary_keys was resolved to 8.1.1, which depends on
      activerecord (~> 4.2.0)

    paper_trail was resolved to 4.0.1, which depends on
      activerecord (< 6.0, >= 3.0)

    rails (< 5.1, >= 5.0.0.beta2) was resolved to 5.0.0.beta2, which depends on
      activerecord (= 5.0.0.beta2)

Running `bundle update` will rebuild your snapshot from scratch, using only
the gems in your Gemfile, which may resolve the conflict.
```

cfis:
Sorry I was mistaken.  Ok, updated the gem file.  And updated the finder methods.  But still a long way to go, patches welcome.  Use the master branch.

cfis:
See #348 and master branch.

PR #348, opened 2016-03-19
Upgrading CPK to Work With Rails 5

codeodor:
Just as an FYI, I'm working on it, so if others are, let's try not to duplicate work. :smile: 

References #332


codeodor:
This will be a slow process as I work through the errors in my app. Some will be CPK, some will be other gems. After I get the app's tests passing, I'll work on the CPK test suite.

cfis:
Sounds good - thanks for tackling this.  We will also update our app, but it likely not for a few more months.

codeodor:
Got my apps tests down to only 6 failures, none of which I think are CPK related.

Unfortunately, my test suite is now running 36+ minutes, and it was certainly not even in double digits of minutes before now. Because of how long it takes to run, I don't know if I'm gonna make a lot of progress in determining if those 6 failures are CPK-related or not.

I don't yet know if the huge performance hit is CPK-related or due to changes in Rails itself.

There are a lot of deprecation notices that I'm considering working on next. 

It's going to be hard to untangle all them though. A ton (maybe all?) run through one part of CPK, but that's not where the deprecation notice is coming from ... it's just calling `#to_a` on the relation, which finally executes it after who knows where it came from, so :sob:

codeodor:
Decided to go ahead and knock out some of the CPK gem test failures. So far:

mysql: `171 runs, 369 assertions, 13 failures, 20 errors, 0 skips` 
postgresql: `171 runs, 364 assertions, 9 failures, 23 errors, 0 skips`
sqlite3: `171 runs, 364 assertions, 9 failures, 23 errors, 0 skips`

codeodor:
I found out today that the issue with performance I was seeing was fixed in `rails/master`, so it's good I don't have to look into that any further!

cfis:
Nice!

codeodor:
TODO: Need to test Rails without CPK to see if it autosaves associations even when there are no changes to the "parent" model.

Our tests expect it, but I had to add in a fake call to `<random_attribute>_will_change!` to make it work.

codeodor:
Progress is slow, but moving forward a bit:

```
mysql: 171 runs, 389 assertions, 9 failures, 13 errors, 0 skips
postgresql: 171 runs, 384 assertions, 5 failures, 16 errors, 0 skips
sqlite3: 171 runs, 384 assertions, 5 failures, 16 errors, 0 skips
```

codeodor:
@cfis, can you help me understand [what this was intended to fix](https://github.com/composite-primary-keys/composite_primary_keys/blob/master/lib/composite_primary_keys/composite_predicates.rb#L11-L27)?

It seems to break when there are more than two columns making up a composite key, so I rewrote it to not do that in this work (just needed to set the return variable from the call to `inject`) since it was not modifying in place). 

However, even with my "fix" it breaks if you are looking up multiple records at a time, because it results in something like this:

`where id1 = 1 and id2 = 2 or id1 = 3 and id2 = 4`  (since it does not group them first)

I want to go back to the original implementation since that is indeed what it should be doing, but I wanted to bring it up in case there's something else that needs to be done to account for what it was intended to do.

codeodor:
Probably stopping for today, but still a bit closer:

```
mysql: 171 runs, 401 assertions, 6 failures, 8 errors, 0 skips
postgresql: 171 runs, 394 assertions, 4 failures, 11 errors, 0 skips
sqlite3: 171 runs, 394 assertions, 4 failures, 11 errors, 0 skips
```

codeodor:
```
mysql: 172 runs, 409 assertions, 5 failures, 5 errors, 0 skips
postgresql: 172 runs, 402 assertions, 3 failures, 8 errors, 0 skips
sqlite3: 172 runs, 402 assertions, 3 failures, 8 errors, 0 skips
```

A tiny bit more progress today. I suspect all but one of those issues are due to:

1. Somewhere we're not splitting the keys into separate clauses (` where id1, id2 = 3, 4` as an example)
2. `#build` is not setting the keys on an association

The last issue is something weird with `#create` which may be related to number 2 above, now that I think about it.

codeodor:
```
mysql: 174 runs, 424 assertions, 5 failures, 0 errors, 0 skips
postgresql: 174 runs, 417 assertions, 3 failures, 3 errors, 0 skips
sqlite3: 174 runs, 417 assertions, 3 failures, 3 errors, 0 skips
```

codeodor:
OK, we're all green now, so at least I suppose we should feel relatively safe in upgrading.

Still need to figure out what's causing the deprecation notices though. That's a tough one, because it happens somewhere with how we build the queries, and I think it never makes its way into the call stackif you were to print the whole thing from the deprecation. :cry:

codeodor:
:warning: Reviewing my app's test failures, I'm seeing a couple that lead me to believe part of CPK is interfering with the "normal" Rails operation of non-composite keys. Just wanted to note this here to hopefully remind me to review it, and as a warning against anyone using this in its current state.

codeodor:
There are still deprecations and warnings, but I found one major source of them today. That one was a LOOOONG time coming. Funny that when I'm just about ready to give up (here and elsewhere), I remember there is one last source I can turn to for answers: `set_trace_func`

codeodor:
:fireworks: :tada: :balloon: All tests pass on the 3 DBMS I can check, all warnings removed or silenced, all deprecation notices fixed. 

So now I feel comfortable in removing the [WIP] tag. 

@cfis I think this is ready to merge after your review.

We'll likely need to update again as Rails releases more betas or RCs, and then the final version, but at least for now anyone who's working on upgrading their app could have something compatible with current the Rails master branch.

cfis:
Awesome work Sammy!  Will cut a 9.0.0.beta gem.

PR #354, opened 2016-04-30
Update #in_batches to work with Rails 5 beta 4

codeodor:


cfis:
Sorry, should have updated this one first.  I think fixed by r357

PR #361, opened 2016-06-20
Bugfix/ar_4.0.x: read from attribute rather than `send` in uniqueness validation

zvkemp:
As corrected in https://github.com/rails/rails/pull/7072


cfis:
Thanks!  Do you need a new gem release?

PR #363, opened 2016-07-13
Create OR predicates in a nicely balanced tree.

### LINKED ISSUE #320: cpk_or_predicate stack level too deep

nathansamson:
The creation is done recursively instead of blindly iteratively.
This has the advantage when generating the actual SQL code the
depth of the tree will be a lot smaller (log_2), and as a result
will not generate a stack level too deep problem.

This fixes #320

### DIFF
--- a/lib/composite_primary_keys/composite_predicates.rb
+++ b/lib/composite_primary_keys/composite_predicates.rb
@@ -8,17 +8,22 @@ module CompositePrimaryKeys
       end
     end
 
-    def cpk_or_predicate(predicates)
+    def cpk_or_predicate(predicates, group = true)
       if predicates.length <= 1
         predicates.first
       else
-        predicates_copy = predicates.dup
-        predicate = predicates_copy.shift
+        split_point = predicates.length / 2
+        predicates_first_half = predicates[0...split_point]
+        predicates_second_half = predicates[split_point..-1]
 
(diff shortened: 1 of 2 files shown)


nathansamson:
If the stack level is 100; and each OR grouping generates 1 entry on the stack you used to be able to query 100 records at a time. Now 2**100 should be possible.

In reality the stacklevel is a bit deeper but each predicate generally creates at least 2 stack entries (as the innermost predicate would be an AND).

I am not sure what the magic failure threshold was for me, but I believe it was around 300. Now it should be closer to 2**300, which is more than the number of atoms in the universe.

CyborgMaster:
This is beautiful!  I have been waiting for a fix to this for over a year.  Lets get this merged ASAP!

cfis:
Thanks!

PR #364, opened 2016-07-22
merge results from INSERT into key

kereyroper:
Sometimes it is useful for only a portion of the primary key to be generated by
the database. Merge the results from the INSERT back into the ActiveRecord
object.

Let me know if there are any concerns with the high-level approach.

Tested in our application with PostgreSQL, which uses INSERT INTO ... RETURNING ...

TODO
- unit tests
- any style cleanup/logic fixes/portability fixes requested (Ruby is not my strongest language at the moment)
- port to newer versions of ActiveRecord

### DIFF
--- a/lib/composite_primary_keys/persistence.rb
+++ b/lib/composite_primary_keys/persistence.rb
@@ -57,5 +57,51 @@ module ActiveRecord
         true
       end
     end
+
+    def save(*)
+      create_or_update
+    rescue ActiveRecord::RecordInvalid
+      false
+    end
+
+    def save!(*)
+      create_or_update || raise(RecordNotSaved)
+    end
+
+    private
+
+    def create_or_update
+      raise ReadOnlyRecord if readonly?
+      result = new_record? ? _create_record : _update_record
+      result != false
+    end
+
+    # Updates the associated record with values matching those of the instance attributes.
+    # Returns the number of affected rows.
+    def _update_record(attribute_names = @attributes.keys)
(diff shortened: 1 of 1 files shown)


cfis:
I think the idea is fine, but seems like some methods were added to the pull request that aren't actually changed from the base Rails request.  So can this be cleaned up to only included changed methods?  Also, would definitely need some tests before merging.

kereyroper:
@cfis You're correct, however, the required changes were to methods that are private, and the internal API was unstable across patch versions within the 4.1 minor version. `_create_record` is named `create_record` before df1df87e for example. `create_or_update` could be subject to a similar renaming.

Let me know if you have a concrete recommendation that would reduce code duplication without increasing risk of incompatibility across patch revisions. It should still be safe to effectively backport fixes from an application-perspective if someone updates composite_primary_keys without updating rails to the revision containing the fix in the base class in most cases.

cfis:
Yes, patching active record is a certainly a pain because every version is different.  Think we just have to live with that fact and not duplicate unchanged methods.  Note that the lastest CPK for the 4.1 branch requires a version that is *after* the _create_record change.

kereyroper:
OK, given that, I can simplify. I guess that's what tests and strict dependency versioning are for. Most likely I'll close this out and port to 4.2/5.0 since 4.1 is EOL, but that will take me a bit to get around to.

cfis:
Going to close this but please reopen if you want to get a change merged.

PR #368, opened 2016-09-29
Fix reset_primary_keys

desofto:
It fixes the problem when AR begins fail with exception `ActiveRecord::UnknownPrimaryKey` on nested model that does not use composite_primary_keys.

### DIFF
--- a/lib/composite_primary_keys/base.rb
+++ b/lib/composite_primary_keys/base.rb
@@ -16,9 +16,8 @@ module ActiveRecord
 
       # Don't like this method name, but its modeled after how AR does it
       def reset_primary_keys
-        if self != base_class
-          self.primary_keys = base_class.primary_keys
-        end
+        return if self == base_class
+        self.primary_keys = base_class.primary_keys || base_class.primary_key
       end
 
       alias_method :primary_key_without_composite_key_support=, :primary_key=


cfis:
Do you have an example of this?  A test case would be great.

desofto:
So sorry, I forgot about this. Here is example:

```ruby
class A < ActiveRecord::Base
  self.primary_keys = [:id_1, :id_2]
  ...
end

class B < A
  ...
end

B.find(1, 1)
=> NameError: uninitialized constant CompositePrimaryKeys::ActiveRecord::FinderMethods::UnknownPrimaryKey
```

desofto:
I cannot reproduce it in Rails console or in test. I've two STI models A and B and do not use compose keys in they:
```ruby
class A < ActiveRecord::Base
end

class B < A
end
```
In controller I left only finder:
```ruby
def show
  B.find(params[:id])
  render text: :ok
end
```

After some time (usually, since second call) finder begins raise exception `Unknown primary key for table companies in model B` until restarting. With this fix it never happens. In console I saw few times this exception, but do not have steps to reproduce.

desofto:
I've found the root of this problem:

https://github.com/flyerhzm/bullet/blob/master/lib/bullet/ext/object.rb#L7

it calls `primary_keys` for class where they are not declared (like in my example), in `primary_keys` it calls `reset_primary_keys` and next `self.primary_keys = base_class.primary_keys` while `base_class.primary_keys == nil`

cfis:
I fixed this by not having the method primary_keys defined if a model is not CPK.  See [master bb6ee75]. Should hopefully fix the issue.

PR #370, opened 2016-11-22
Don't nest PK twice when looking up id, fixes #319

### LINKED ISSUE #319: Problem with tables that have an ID column
Hi,
Recently I tried to update to version 8.1.1 because I faced #291.
But, unfortunately, a lot of my tests broke because I have ID columns in some tables.

For example, considering the following models:

``` ruby
class Person < ActiveRecord::Base
  self.primary_keys = :id, :company_id

  has_many :documents, class_name: 'PersonDocument', primary_key: [:id, :company_id],
                   foreign_key: [:person_id, :company_id]

  has_many :wtf_documents, ->(record) { where(person_id: record.id[0][0],
                                         company_id: record.company_id) },
           class_name: 'PersonDocument',
           foreign_key: ['person_id', 'company_id']

end

class PersonDocument < ActiveRecord::Base
  self.primary_keys = :id, :company_id
end
```

With tables:

``` ruby
class CreatePeople < ActiveRecord::Migration
  def change
    create_table :people do |t|
      t.string :name
      t.references :company, index: true
      t.timestamps null: false
    end
  end
end

class CreatePersonDocuments < ActiveRecord::Migration
  def change
    create_table :person_documents do |t|
      t.string :document_number
      t.integer :person_id, index: true
      t.integer :company_id, index: true
      t.timestamps null: false
    end
  end
end
```

After creating some records:

``` ruby
Person.create(company_id: 1, name: 'Person name')
PersonDocument.create(person_id: 1, company_id:1, document_number: '4as65a4s654')
```

If I try to get the object id, this is my result:

``` ruby
pe = Person.first
#SELECT  "people".* FROM "people"  ORDER BY "people"."company_id" ASC LIMIT 1
# => #<Person id: [1, 1], name: "Person name", company_id: 1, 
# created_at: "2015-08-20 00:21:33", updated_at: "2015-08-20 00:21:33">
pe.id
# => [[1, 1], 1] 
```

So, if I try to get the has_many association this is my result:

``` ruby
pe.documents
# SELECT "person_documents".* FROM "person_documents" WHERE 
# ("person_documents"."person_id" = NULL AND "person_documents"."company_id" = 1)
# => #<ActiveRecord::Associations::CollectionProxy []>
```

Even though I try to pass the ids of  has_many association directly, the problem persist:

``` ruby
pe.wtf_documents
# SELECT "person_documents".* FROM "person_documents" WHERE 
#("person_documents"."person_id" = NULL AND "person_documents"."company_id" = 1) AND 
# "person_documents"."person_id" = $1 AND "person_documents"."company_id" = $2  
#[["person_id", 1], ["company_id", 1]]
# => #<ActiveRecord::Associations::CollectionProxy []>
```

But, as you can see in this case, duplicates the 'where' clause.

On the other hand, none of these problems occurs if table no have an ID column. In this case, the object id is returned as expected, and associations are loaded correctly:

``` ruby
wi = WithoutIdPerson.first
# SELECT  "without_id_people".* FROM "without_id_people"  ORDER BY 
#"without_id_people"."company_id" ASC LIMIT 1
# => #<WithoutIdPerson my_pk: 1, name: "No ID Person", company_id: 1,
# created_at: "2015-08-20 00:47:12", updated_at: "2015-08-20 00:47:12"> 
wi.id
# => [1, 1] 
```

For now, I went back to version 8.0.0 with Rails 4.2.0 and did a "workaround" to #291... 
Any help is very appreciated, thanks.

kereyroper:
For tables that have ID column and use composite primary keys, the id attribute
is already overridden to fetch multiple attributes. This change prevents
nesting the entire primary key in the first element for invocations of id,
matching the behavior in the ar_4.1.x branch.

### DIFF
--- a/lib/composite_primary_keys/attribute_methods/read.rb
+++ b/lib/composite_primary_keys/attribute_methods/read.rb
@@ -7,7 +7,8 @@ module ActiveRecord
           _read_attribute(attr_name, &block)
         else
           name = attr_name.to_s
-          name = self.class.primary_key if name == 'id'.freeze
+          # CPK
+          name = self.class.primary_key if name == 'id'.freeze && !composite?
           _read_attribute(name, &block)
         end
       end
--- a/test/fixtures/db_definitions/postgresql.sql
+++ b/test/fixtures/db_definitions/postgresql.sql
@@ -218,3 +218,12 @@ create table employees_groups (
   employee_id int not null,
   group_id int not null
 );
+
+create table pk_called_ids (
+    id serial not null,
+    reference_code    int         not null,
+    code_label        varchar(50) default null,
+    abbreviation      varchar(50) default null,
+    description       varchar(50) default null,
+    primary key (id, reference_code)
+);
--- a/test/fixtures/pk_called_id.rb
+++ b/test/fixtures/pk_called_id.rb
@@ -1,0 +1,5 @@
+class PkCalledId < ActiveRecord::Base
+  self.primary_keys = :id, :reference_code
+
+  validates_presence_of :reference_code, :code_label, :abbreviation
+end
--- a/test/fixtures/pk_called_ids.yml
+++ b/test/fixtures/pk_called_ids.yml
@@ -1,0 +1,11 @@
+name_prefix_mr:
+  id: 1
+  reference_code: 1
+  code_label: MR
+  abbreviation: Mr
+
+name_prefix_mrs:
+  id: 2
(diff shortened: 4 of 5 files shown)


kereyroper:
Tests need to be ported to other databases. It appears that tests may also have been broken before this change for other databases, since an unrelated MySQL test is failing.

cfis:
Thanks!

PR #372, opened 2016-12-16
Fix `establish_connection` not working with string

syndbg:
When passing a string `spec` such as `"development"`,
it needs to be casted to symbol `:development`,
due to the `resolver` expecting a Hash or a symbol.

Example usage:

```
class ReportingApplicationRecord < ActiveRecord::Base
  self.abstract_class = true

  # BROKEN - And now fixed
  establish_connection('reporting_db')
  # GOOD - Still working
  establish_connection(:reporting_db)
end
```

@cfis @codeodor

### DIFF
--- a/lib/composite_primary_keys/connection_adapters/abstract/connection_specification_changes.rb
+++ b/lib/composite_primary_keys/connection_adapters/abstract/connection_specification_changes.rb
@@ -10,6 +10,13 @@ module ActiveRecord
     end
 
     def self.establish_connection(spec = nil)
+      # NOTE: When passing a string `spec` such as `"development"`,
+      # it needs to be casted to symbol `:development`,
+      # due to the `resolver` expecting a Hash or a symbol.
+      if spec.is_a?(String)
+        spec.to_sym!
+      end
+
       spec     ||= ActiveRecord::ConnectionHandling::DEFAULT_ENV.call.to_sym
       resolver =   ConnectionAdapters::ConnectionSpecification::Resolver.new configurations
       spec     =   resolver.spec(spec, self == Base ? "primary" : name)


codeodor:
Hey thanks @syndbg!

When CPK has to modify a Rails method, so far as I've been able to tell, it tries to stay closely aligned with the original method definition in Rails, and only modify what needs modifying for CPK to work.

When I tried it in a non-CPK app in Rails 4.2.7.1, I got this message:
`DEPRECATION WARNING: Passing a string to ActiveRecord::Base.establish_connection for a configuration lookup is deprecated, please pass a symbol (:development) instead.`

And judging by (what I think is) [the original method in Rails](https://github.com/rails/rails/blob/v5.0.0/activerecord/lib/active_record/connection_handling.rb#L47-L62), I think they removed this behavior.

Though I did not understand why they chose to do that, after reading [this comment in the deprecation notice code](https://github.com/rails/rails/blob/v4.2.7.1/activerecord/lib/active_record/connection_adapters/connection_specification.rb#L219-L235):

```ruby
# Rails has historically accepted a string to mean either
# an environment key or a URL spec, so we have deprecated
# this ambiguous behaviour and in the future this function
# can be removed in favor of resolve_url_connection.
```

I think I understand why. 

It would be my preference that we stick to Rails' behavior here, and not deviate any more than we need to to cover composite primary key functionality.

syndbg:
@codeodor I see, then does it make sense to throw a specific error to guide the developer to pass a symbol instead of string? The error I initially received "Adepter not specified" was very lacking.

codeodor:
I can definitely see value in that, but I think it would be more appropriate in Rails itself than in CPK.

(Do keep in mind that it's been since 5.0 was released since I looked at this, so if you disagree, or if you propose it there and they do, I am certainly open to being persuaded.)

codeodor:
(Also, I don't (think I) have any authority here. I just happened to need to upgrade my app and have committed my work to the project.  @cfis may have totally different ideas.)

cfis:
Thanks for the report!  I do agree with codeodor that this is an ActiveRecord issue.  CPK doesn't override the connection logic, so isn't an appropriate place to fix this issue.

Issue #355, opened 2016-05-01
CPK does not load records when passed in as attributes before save


codeodor:
I don't have time to investigate this right now, but I thought I'd provide a general description for later, and also to solicit any feedback from people with similar issues.

Basically, I have this setup:

``` ruby
class Apartment < ActiveRecord::Base
   has_many :rooms, ->{ includes(:tenant) }
end

class ApartmentRoom < ActiveRecord::Base
  belongs_to :tenant, foreign_key: [:tenant_id, :apartment_id]
end
```

Then, we create records like:

``` ruby
apartment.update_attributes(params[:apartment]) # which includes nested attributes for ApartmentRoom
```

If we call `apartment.rooms` after simply loading an `Apartment`, all is well: the `tenant` is there. But if we try `apartment.rooms.first.tenant` after the call to `update_attributes`, `tenant` is `nil` for all rooms, and it doesn't even look like it tries to load.

For now I just have code to override it like this:

``` ruby
class ApartmentRoom < ActiveRecord::Base
  belongs_to :tenant, foreign_key: [:tenant_id, :apartment_id]

  def tenant
    super || Tenant.find_by(...)
  end
end
```

But I think CPK might need a fix for this issue.

Side note: this is running on the Rails 4 branch. I haven't tried it on the Rails 5 stuff yet.

cfis:
Did you ever get chance to check this on Rails 5?  Mind writing a quick test case?

codeodor:
@cfis I *think* I found my code and tried it out on Rails 5, and there was no error.

I think a test case would be valuable to be sure it's gone and doesn't come back, but I don't think I'll have time for at least the next 2 weeks. 

Mind keeping this open a bit longer until I can do that?

cfis:
Sure, no problem.

codeodor:
Ok, I spent a couple of hours on this today and I couldn't even get it to happen on the 4.2 branch.

I think there must have been an issue, but because it happened so long ago, I lost whatever context I had and probably missed something when I wrote up the issue.

So I think we can close this, and if we run into it in the future, we can just fix it then.

cfis:
Ok - cool.

PR #375, opened 2017-01-08
Development updates

codeodor:
When working on #355 I made a change to:

1) allow developers to run the tests without installing gems to work with DBs they don't have
2) restore the tests that had been commented out

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ debug.log
 local/*.rb
 pkg
 db
+.bundle/config
 
 /doc
 /.idea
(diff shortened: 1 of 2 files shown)


cfis:
Thanks. FYI on the tests, you can use bundle --without.  So if you want to deal with just postgresql:

bundle install --without="oracle msqserver sqlite3 msql" 

Meant to update the readme with that information.

codeodor:
Thanks! That's how I did it, but Bundler created a config folder, so that's why I `.gitignore`d it. :)

cfis:
Ah, makes sense.

Charlie

PR #377, opened 2017-01-11
extend approach from #344 to fix autosave for has_one associations as well

cfinucane:
See #344 for details.


cfis:
Thanks for the patch.  Can you add  a test please?

Any interesting in porting this to master for Rails 5 (#344 hasn't been ported either I notice).

PR #380, opened 2017-01-22
Associations with inverse fix

reist:
Fixing bug in #344 and fixing&adding test for #377

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ db
 
 test/connections/databases.yml
 Gemfile.lock
+vendor/bundle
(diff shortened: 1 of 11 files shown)


cfis:
Thanks for the pull request.  Couple things:

* Can you keep the original source line from rails, just comment it out and put a comment above that says "CPK."  Makes porting in the future easier (and every major Rails version so far has required porting)

* Any chance the existing tables can be used for tests?  Trying to have less tables, not more.  If we do have to add tables then they need to be added for all database adapters and tested across them.

reist:
Done.
When I added the new table, I did it on all the open DBs and tested on them (mysql, pg, sqlite3), but I managed to do the same with the existing tables. I still had to modify `employees`, as I needed a has_one with a composite `primary_key` I could stick an `inverse_of` on. Also, I prefer to modify a non-referencing field of that table, so the test will have less moving parts.
I also added the modification to sqlserver and oracle, as they had the table defined. Hope I did it correctly.

cfis:
Cool thanks!

PR #381, opened 2017-01-23
Fix TestIds#test_set_ids_string

reist:
The one failing test left on sqlite3 was a mistake in copying pk_called_ids DDL.

### DIFF
--- a/test/fixtures/db_definitions/sqlite.sql
+++ b/test/fixtures/db_definitions/sqlite.sql
@@ -206,7 +206,7 @@ create table employees_groups (
 );
 
 create table pk_called_ids (
-    id serial not null,
+    id integer not null,
     reference_code    int         not null,
     code_label        varchar(50) default null,
     abbreviation      varchar(50) default null,


cfis:
Thanks.  There are also some delete_all failures on sqllite but I don't see a easy way to fix them.

reist:
Which would those be? I didn't see any commented out tests on the branch, and everything now passes...except on Ruby 1.9. That can also be a simple fix in the gemspec.

cfis:
Oh sorry, that's on Rails 5.0 so on master branch.

Its this use case:

MyTable.joins(:another_table).delete_all

Any chance you want to port this stuff to the master branch?

reist:
That's the plan, once I finish working on the code that required these fixes in the first place.

If the delete_all issue has an existing test, I don't see it failing on the master branch on either mysql, pg or sqlite.

reist:
Possibly, it's an issue with an old sqlite3 version. The specs of the instance Travis is running are ancient. I'm running 3.16.1 locally.

Issue #385, opened 2017-02-21
association scope cache return wrong record . / Rails4.2


ttwo32:
I faced to [#291](https://github.com/composite-primary-keys/composite_primary_keys/issues/291).

I have a same model that have mulitiple PK, but not include id column.
Please see the below.

```
class Department < ActiveRecord::Base
self.primary_key = :aid, :bid
end
class Employee < ActiveRecord::Base
belongs_to :department, foreign_key: [:aid, :bid]
end
```

the tables(these are same #291):

```
create_table :employees do |t|
  t.integer :aid, null:false
  t.integer :bid, null:false
  t.string :name

  t.timestamps null: false
end

create_table :departments, id: false do |t|
  t.integer :aid, null:false
  t.integer :bid, null:false
  t.string :deptname

  t.timestamps null: false
end
```

enviroments:

```
Rails: 4.2.7
DB: SQLite
ActiveRecord: 5.2
```

Look at the below.
Association return a first time record.
That was wrong .

```
[1] pry(main)> Department.create!(aid:1, bid:2, deptname: 'd1')
   (0.1ms)  begin transaction
  SQL (0.5ms)  INSERT INTO "departments" ("aid", "bid", "deptname", "created_at", "updated_at") VALUES (?, ?, ?, ?, ?)  [["aid", 1], ["bid", 2], ["deptname", "d1"], ["created_at", "2017-02-21 15:23:31 UTC"], ["updated_at", "2017-02-21 15:23:31 UTC"]]
   (1.9ms)  commit transaction
=> #<Department:0x007fc03feee6a0 aid: 1, bid: 2, deptname: "d1", created_at: "2017-02-21 15:23:31 UTC", updated_at: "2017-02-21 15:23:31 UTC">
[2] pry(main)> Department.create!(aid:3, bid:4, deptname: 'd2')
   (0.2ms)  begin transaction
  SQL (0.3ms)  INSERT INTO "departments" ("aid", "bid", "deptname", "created_at", "updated_at") VALUES (?, ?, ?, ?, ?)  [["aid", 3], ["bid", 4], ["deptname", "d2"], ["created_at", "2017-02-21 15:23:40 UTC"], ["updated_at", "2017-02-21 15:23:40 UTC"]]
   (1.9ms)  commit transaction
[...]

PR #386, opened 2017-02-21
Query cache with bind params

ttwo32:
see detail #385 .


ttwo32:
Woops!! I had mistaken. Sorry.

Issue #383, opened 2017-01-24
activerecord-sqlserver-adapter  OFFSET issue


cyberbruce:
I am receiving an error using
- Rails 5.0.1
- activerecord-sqlserver-adapter (5.0.1)
- tiny_tds (1.1.0)
- composite_primary_keys (9.0.6)

Error 

> TinyTds::Error: Incorrect syntax near 'OFFSET'

I get this after creating a new database during "rails db:environment:set"  command
I also get it during other association calls. 

Are you aware of this issue?   I can post more details, if not.   

Thank you for you time and efforts.   We use this gem quite often, and it has served us well over the years.  I'm really looking forward to having this working in Rails 5.

jbwl:
Same error with the same versions that cyberbruce listed.

pinak1180:
same error with same specs with me as well

oriolbcn:
had the same problem today, check the PR for a solution :)

PR #388, opened 2017-02-23
fix incorrect case statement to get PK for SQLServer

oriolbcn:
Seems like the branches of the case statement are inverted, fixes https://github.com/composite-primary-keys/composite_primary_keys/issues/383


cfis:
Thanks!

PR #391, opened 2017-03-02
Override ActiveRecord::Associations::CollectionAssociation#ids_reader

pustomytnyk:


cfis:
Ok, thanks for the fix!

PR #415, opened 2017-10-03
Fix Paper Trail compatibility

### LINKED ISSUE #412: id is being incorrectly marked as changed
On a model that does **not** set multiple columns as the primary key and has associations updated with `accepts_nested_attributes_for`, the `:id` is marked as having changed by:

https://github.com/composite-primary-keys/composite_primary_keys/blob/master/lib/composite_primary_keys/nested_attributes.rb#L63

This causes Rails to falsely report that `:id` has changed when calling dirty methods like `changed`. Given the code, it looks like this may also happen on models that use multiple columns as their primary key. I haven't tested that case.

seanlinsley:
This fixes #412 by removing `*_will_change!` calls added in #348 via https://github.com/composite-primary-keys/composite_primary_keys/commit/dc0cc456be228748cc0416d7d19d477744dee97a.

Originally `update_attribute` in [Rails 4.2](https://github.com/rails/rails/blob/v4.2.0/activerecord/lib/active_record/persistence.rb#L239) called save no matter if changes existed. [Rails 5.0](https://github.com/rails/rails/blob/v5.0.0/activerecord/lib/active_record/persistence.rb#L255) updated it to only save if `changed?`, and [Rails 5.1](https://github.com/rails/rails/blob/v5.1.0/activerecord/lib/active_record/persistence.rb#L270) updated it to call a new method, `has_changes_to_save?`. The problem is these methods don't return true if only an association was updated, even if the call to save would actually end up committing changes.

This is a change in Rails behavior that I'm sure isn't considered a bug by Rails core team, so we shouldn't introduce a bug into Paper Trail by attempting to patch it. This commit removes the patch, and updates the failing tests to instead call `update_attributes`, which always calls `save`.

### DIFF
--- a/lib/composite_primary_keys/nested_attributes.rb
+++ b/lib/composite_primary_keys/nested_attributes.rb
@@ -59,9 +59,7 @@ module ActiveRecord
           attributes = attributes.with_indifferent_access
           if attributes['id'].blank?
             unless reject_new_record?(association_name, attributes)
-              new_record = association.build(attributes.except(*UNASSIGNABLE_KEYS))
-              public_send("#{self.attributes.keys.first}_will_change!") # TODO CPK: Note: Is this a bug in Rails? We need to somehow set the parent model as changed even when changes are only adding new records to an association. This is not the way to do it, but I am not sure how it should be done.
-              new_record
+              association.build(attributes.except(*UNASSIGNABLE_KEYS))
             end
           elsif existing_record = cpk_detect_record(attributes['id'], existing_records)
             unless call_reject_if(association_name, attributes)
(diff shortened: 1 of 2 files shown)


seanlinsley:
The two test failures exist on master.

seanlinsley:
This should be backported to the 9.x line.

cfis:
Thanks!

seanlinsley:
Happy to help 😄 

I'd appreciate if this could get backported to 9.x, since the bug was introduced there (and we're still using 9.x).

I'd be happy to open a new PR for the `ar_5.0.x` branch if you want to be able to handle it via the GitHub UI.

cfis:
Sure, that works.

seanlinsley:
New PR: #416

PR #417, opened 2017-10-11
Fix a quoting error which occur in rare cases

supercaracal:
[Rails 5.1 upgrade — TypeError: can’t quote Hash](https://medium.com/@vvangemert/rails-5-1-upgrade-typeerror-cant-quote-hash-840a093a2a3f)

> We discovered the problem occurred when using a attribute with the same name as the table.

```
/path/to/project/.bundle/ruby/2.4.0/gems/activerecord-5.1.4/lib/active_record/connection_adapters/abstract/quoting.rb:196:in `_quote': can't quote Hash (TypeError) 
  from /path/to/project/.bundle/ruby/2.4.0/gems/activerecord-5.1.4/lib/active_record/connection_adapters/abstract/quoting.rb:25:in `quote'
  from /path/to/project/.bundle/ruby/2.4.0/gems/activerecord-5.1.4/lib/active_record/statement_cache.rb:52:in `block in sql_for'
  from /path/to/project/.bundle/ruby/2.4.0/gems/activerecord-5.1.4/lib/active_record/statement_cache.rb:52:in `each'
  from /path/to/project/.bundle/ruby/2.4.0/gems/activerecord-5.1.4/lib/active_record/statement_cache.rb:52:in `sql_for'
  from /path/to/project/.bundle/ruby/2.4.0/gems/activerecord-5.1.4/lib/active_record/statement_cache.rb:105:in `execute'
  from /path/to/project/.bundle/ruby/2.4.0/gems/activerecord-5.1.4/lib/active_record/associations/singular_association.rb:50:in `find_target'
  from /path/to/project/.bundle/ruby/2.4.0/gems/activerecord-5.1.4/lib/active_record/associations/association.rb:157:in `load_target'
  from /path/to/project/.bundle/ruby/2.4.0/gems/activerecord-5.1.4/lib/active_record/associations/association.rb:53:in `reload'
  from /path/to/project/.bundle/ruby/2.4.0/gems/activerecord-5.1.4/lib/active_record/associations/singular_association.rb:7:in `reader'
```
see also:
* [ActiveRecord::ConnectionAdapters::Quoting#_quote](https://github.com/rails/rails/blob/v5.1.4/activerecord/lib/active_record/connection_adapters/abstract/quoting.rb#L196)
* [ActiveRecord::Associations::AssociationScope#last_chain_scope](https://github.com/rails/rails/blob/v5.1.4/activerecord/lib/active_record/associations/association_scope.rb#L63)
* [ActiveRecord::Associations::AssociationScope#apply_scope](https://github.com/rails/rails/blob/v5.1.4/activerecord/lib/active_record/associations/association_scope.rb#L166)
* [Specify `table.name` only when `scope.table` and `table` are different](https://github.com/rails/rails/pull/29058)

### DIFF
--- a/lib/composite_primary_keys/associations/association_scope.rb
+++ b/lib/composite_primary_keys/associations/association_scope.rb
@@ -37,15 +37,11 @@ module ActiveRecord
           hash
         end
 
-        if scope.table == table
-          scope = scope.where(joins)
-        else
-          scope = scope.where(table.name => joins)
-        end
+        scope = cpk_apply_scope(scope, table, joins)
 
         if reflection.type
           polymorphic_type = transform_value(owner.class.base_class.name)
-          scope = scope.where(table.name => { reflection.type => polymorphic_type })
+          scope = cpk_apply_scope(scope, table, { reflection.type => polymorphic_type })
         end
 
         scope
@@ -62,10 +58,20 @@ module ActiveRecord
 
         if reflection.type
           value = transform_value(next_reflection.klass.base_class.name)
-          scope = scope.where(table.name => { reflection.type => value })
+          scope = cpk_apply_scope(scope, table, { reflection.type => value })
         end
 
         scope = scope.joins(join(foreign_table, constraint))
+      end
+
+      private
+
+      def cpk_apply_scope(scope, table, joins)
+        if scope.table == table
+          scope.where(joins)
+        else
+          scope.where(table.name => joins)
+        end
       end
     end
   end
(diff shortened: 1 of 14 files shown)


cfis:
Nice find, I ran into the same issue and was investigating.

However, let's not fix it this way.  The code you overrode is from Rails 5.  It should first be updated from Rails 5.1.  Then it should be fixed.

Note rails 5.1 changed the code to use a method called apply_scope, which is the similar to the cpk_apply_scope the patch adds.  So cp_apply_scope isn't needed.

supercaracal:
@cfis Is it mean we should have not partial optimization but to merge from Rails 5.1 perfectly?

supercaracal:
@cfis I understood d9b09e1. Sorry.

PR #426, opened 2018-02-14
Update for rails 5.1.5 support

tomhughes:

PR #429, opened 2018-02-25
Version 10.0.3

edsharp:
Following @tomhughes  update in https://github.com/composite-primary-keys/composite_primary_keys/pull/426 would it be possible to release this as 10.0.3?

Many thanks

Ed


cfis:
Remember, the gem requirements also need to be updated (which I did).

cfis:
Thanks!

Issue #365, opened 2016-07-28
Primary keys not set after create


kirillsalykin:
```
class Rec < ApplicationRecord
  self.primary_keys = [:seq, :status]
end

rec = Rec.create
rec.id # => [nil, nil]
```

`req` - is sequential integer, eg database sets default value.

Rails uses ||= for setting generated ids 
```
self.id ||= new_id if self.class.primary_key
```
https://github.com/rails/rails/blob/master/activerecord/lib/active_record/persistence.rb#L560

Id for transient record is [nil, nil], thus it will be never overwritten with values from db.

cfis:
Yeah, I think you'll need to do this by hand in your application code.  I guess CPK could check each primary key's type, determine if its a serial, and if yes fill in the value.  But this issue has never come up before so I'm guessing this is fairly unusual case.

PR #431, opened 2018-03-16
Set Rails "id" column value on insert

anttipitkanen:
I noticed this old issue https://github.com/composite-primary-keys/composite_primary_keys/issues/365 and decided to give this a stab as our code base would really need this functionality.

It's admittedly not the most common use case, but as we are sharding an existing database with bunch of Rails style `id` columns it would be really convenient to get the `id` back on insert. In our use case there's a composite primary key `(id, <shard key>)` in the table so you can't refetch the record easily to obtain the `id` afterwards like done here: https://github.com/rails/rails/issues/17605

I tried to keep the impact of this PR minimal by only setting the special `id` column value on insert if it's not set yet. I guess there's other use cases where you have several columns that get populated by functions on the database side, but I'm not sure if ActiveRecord would be able to return them (like in the above rails issue)


cfis:
Sorry for the delay.  This code looks correct to me, but it seems like such a narrow use case.  You'd have to have a serial column named id.  Another column type, like uuid, wouldn't work.

I hate for you to have done this work and not merge it, but this one does seem more application specific to me.  Maybe better for you to just monkey patch CPK in your Rails app?

anttipitkanen:
Admittedly it's narrow, but then again Rails must be quite a huge portion ActiveRecord usage.

I'm not sure what the problem would be with uuid type of column, can you explain in more detail? This change should only make the assumption that column with Rails style name `id` is autogenerated and the returned value should be set to the model. I don't see how an uuid generated in the same way of auto incrementing integer would break here?

anttipitkanen:
But yeah, it's feasible to monkey patch this in the app itself too.

cfis:
I though that UUID columns did not autogenerate new ids, but now that I looked at the postgresql doc they in fact do.

mattgrande:
I could also use this, so I would appreciate getting it in :)

cfis:
I think as written, this isn't worth merging into CPK.  I could get behind an alternative approach though which would be update _insert_record to return the new id, thus not having to change _create_record. Not sure how _insert_record figures out what the new key is, but it the change would hopefully just do the same thing but loop over each primary key field.

codeodor:
I'm trying to understand the use case. Is the auto-generated value for your ID column not unique?

mattgrande:
@codeodor - For me, no. I have a combination of ID + Version (the table is insert-only, and changes are recorded by bumping a version number).

codeodor:
Ah, thank you for helping me understand!

hiradyazdan:
you may have another column that is not autogenerated uuid as a surrogate key, and then need to pass it on to associated tables down the line. How would you patch a transient record when there's no id until it's created?

PR #433, opened 2018-03-28
Set inversion when using preload Rails 5.1.x

letrungkien:
Thanks for making this awesome gem.
We had N+1 problem when upgrading to Rails 5.1.5 using composite-primary-keys v10.0.3

Assuming we have `Country` and `City` model and `Country` has_many `City`.
```ruby
countries = Country.preload(:cities)
city = countries.first.cities.first
city.association(:country).loaded? #=> false
city.country # will trigger call to DB
```

Not sure why inversion is not set. The difference lies in these 2 lines: https://github.com/rails/rails/blob/v5.1.5/activerecord/lib/active_record/associations/preloader/association.rb#L63-L64

I noticed that tests failed due to the recent update in `mysql2` from `0.4.10` to `0.5.0` and made a separate PR to address the issue: #434

### DIFF
--- a/lib/composite_primary_keys/associations/preloader/association.rb
+++ b/lib/composite_primary_keys/associations/preloader/association.rb
@@ -43,6 +43,8 @@ module ActiveRecord
               records.each do |record, owner_key|
                 owners_map[owner_key].each do |owner|
                   records_by_owner[owner] << record
+                  association = owner.association(reflection.name)
+                  association.set_inverse_instance(record)
                 end
               end
             end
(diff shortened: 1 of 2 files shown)


saveav:
👍 should good to go, no?

cfis:
Cool - thanks!

letrungkien:
@cfis Thank you!
We're upgrading our Rails to 5.2 and it will be great if you could add these changes to version 11.x.

letrungkien:
@cfis sorry to trouble you.
I checked that v11.x already handled the issue here:
https://github.com/composite-primary-keys/composite_primary_keys/blob/v11.0.3/lib/composite_primary_keys/associations/preloader/association.rb#L106-L107

cfis:
Great. I'll add in the test you provided also for the master branch.

PR #434, opened 2018-03-28
Lock mysql2 to ~> v0.4.0 for Rails 5.1.x

letrungkien:
I noticed it will fail if checking out from `ar_5.1.x` because `mysql2` just updated `0.5.0` recently

Example failed build: https://travis-ci.org/composite-primary-keys/composite_primary_keys/jobs/359215301


cfis:
Thanks!

PR #442, opened 2018-04-14
Fixes for Rails 5.2. Exception on update and delete

erkie:
Commit 1: Update `_update_record` fixes the fact that the method signature changed. However, due to the change of method parameters I ended up using `constraints[primary_key]` which I'm not at all sure works
Commit 2: My first ever encounter with AREL. I have no idea if this accurately reflects what the method intent is, but my tests passed. It also uses the `constraints[primary_key]` construction which seems brittle.


Szeliga:
Hey, any plan on fixing the tests on this one? Or maybe passing it over to someone else?

erkie:
@Szeliga I was lucky enough to not rely on `composite_primary_keys` enough, so I could just rip it out entirely from my project so this PR is not urgent for me.

But I'd be more than happy to finish this PR, it's literally just waiting on help from the maintainers so we can get this merged. But they seem to be fairly non-existent...

Szeliga:
@erkie no rush, we are not depending on it that much either and I was considering removing it altogether, but several tests started to fail because without a primary key equality checking in ActiveRecord models doesn't work.

PR #448, opened 2018-05-30
ActiveRecord 5.1.x fixes

RISCfuture:
* Update implementation of `_update_record`  method in `Persistence` to match new signature in 5.1.x
* Rename `_update_record` to `_update_row` in `OptimisticLocking`
* Revert patches to `AttributeSet::Builder` as they are no longer necessary

All tests pass.

Not sure if the `_update_record` method needs to be modified to handle symbols as primary keys, or if they will all be normalized to strings.


joevandyk:
Does this also fix 5.0.x?

RISCfuture:
Not sure. And this isn't purporting to be a comprehensive list of necessary fixes for 5.1 compatibility. This is merely the monkey-patches I had to make to get my project to pass specs on 5.1.

gschorkopf:
This is great work and will really help unblock a project I'm working on! Could we get an owner with merge-access on this library to take a look? Maybe @cfis ? ❤️

cfis:
So this change is not compatible with Active Record 5.1.0 through 5.1.6?

If that is the case, then I think the gemspec needs to be updated to require ActiveRecord 5.1.6 or higher and the gem file.

PR #449, opened 2018-06-01
Add friendly error for incomplete args

jamesBrennan:
One of the things I've found frustrating about working with `composite_primary_keys` is that the errors it generates tend to be cryptic and not helpful for figuring out what actually went wrong. This PR seeks to begin addressing that.

I've added `CompositePrimaryKeys::IncompleteArgumentsError`, which is raised when `.find()` is called with fewer arguments than are expected.

```ruby
# Given a class with composite primary keys
class Department < ActiveRecord::Base
  self.primary_keys = :department_id, :location_id
end

Department.find(1)
# ...
# CompositePrimaryKeys::IncompleteArgumentsError (Can not map arguments to composite primary key)
#   args: 1
#   primary key: ["department_id", "location_id"]
#   class: Department
```

Also covered:
```ruby
# These all raise CompositePrimaryKeys::IncompleteArgumentsError
Department.find([1,nil])
Department.find([1,1], 2)
Department.find('1,1', '2')
Department.find('1,')
```

### DIFF
--- a/test/test_find.rb
+++ b/test/test_find.rb
@@ -120,6 +120,44 @@ class TestFind < ActiveSupport::TestCase
     assert_equal([1,3], ref_code.id)
   end
 
+  def test_find_with_incomplete_id
+    begin
+      ReferenceCode.find(1)
+    rescue StandardError => e
+      assert_kind_of(CompositePrimaryKeys::IncompleteArgumentsError, e)
+      assert_match('primary_key: ["reference_type_id", "reference_code"]', e.message)
+      assert_match('args: 1', e.message)
+      assert_match('class: ReferenceCode', e.message)
+    end
+  end
+
+  def test_find_some_with_incomplete_id
+    begin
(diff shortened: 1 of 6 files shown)


test/test_find.rb
```diff
@@ -120,6 +120,44 @@ def test_find_one_with_params_id
...
+    rescue StandardError => e
+      assert_kind_of(CompositePrimaryKeys::IncompleteArgumentsError, e)
+      assert_match('args: ["1"]', e.message)
+    end
+  end
+
+  def test_find_with_nil_value
+    begin
+      ReferenceCode.find([1,nil])
+    rescue StandardError => e
```
mberlanda: I would avoid to rescue `StandardError`. You may wanting to catch only ActobeRecord relates exceptions

cfis:
Sorry not to have responded to this commit. I'm going to close it - I'd rather not introduce new types of exceptions not found in ActiveRecord.

PR #451, opened 2018-06-06
Convert batches to arrays before continuing loop

CyborgMaster:
This is the same concept as used in the main rails code here:

https://github.com/rails/rails/blob/v4.2.10/activerecord/lib/active_record/relation/batches.rb#L128

Before looping to the next iteration of `find_in_batches` we populate the relation to an array.  Without this the `while records.any?` check at the top of the loop ends up performing an extra count query, and at worse it was crashing for me (but that might have been an interaction with another gem I'm using)


cfis:
Thanks!

Issue #459, opened 2018-08-10
Error on calling private method aliases with rails > 5.2


jklimke:
Hi there. I noticed the following error after upgrading to 11.0.3.

It seems that recent changes using a private method.

```
Error
private method `aliases' called for #<ActiveRecord::Associations::JoinDependency:0x000000131b15f8>

Stacktrace
private method `aliases' called for #<ActiveRecord::Associations::JoinDependency:0x000000131b15f8> (NoMethodError)
NoMethodError:
private method `aliases' called for #<ActiveRecord::Associations::JoinDependency:0x000000131b15f8>
/home/jenkins/.rvm/gems/ruby-2.3.3/gems/composite_primary_keys-11.0.3/lib/composite_primary_keys/relation/finder_methods.rb:21:in `apply_join_dependency'
./spec/models/real_attribute_spec.rb:18:in `block (4 levels) in <top (required)>'
/home/jenkins/.rvm/gems/ruby-2.3.3/gems/composite_primary_keys-11.0.3/lib/composite_primary_keys/relation/batches.rb:40:in `block in in_batches'
/home/jenkins/.rvm/gems/ruby-2.3.3/gems/composite_primary_keys-11.0.3/lib/composite_primary_keys/relation/batches.rb:18:in `loop'
/home/jenkins/.rvm/gems/ruby-2.3.3/gems/composite_primary_keys-11.0.3/lib/composite_primary_keys/relation/batches.rb:18:in `in_batches'
./spec/models/real_attribute_spec.rb:17:in `block (3 levels) in <top (required)>'    
```

kingdonb:
I just encountered this too.  The code that triggered the issue is a "where" on an association that does "eager_load" through some composite primary key associations.  It worked fine against Rails 5.2.0, but last night I tried upgrading to 5.2.1 and hit this error exactly.

I think I can understand the issue, I didn't even consider that it had something to do with my composite primary keys.  I guess I should get used to this...

Issue #444, opened 2018-04-17
delete_all ActiveRecord instance method does not work properly anymore


Victorcorcos:
On the latest version of this gem, the `delete_all` instance method is deleting everything on the model even if some restrictions are applied before the deletion. This bug particularly happens when a model is joined with another one before the deletion.

For example, assume that a random model (`Asset`) `belongs_to` other random model (`Project`).
The `Asset` model has `5` records.

Let see what happens with the following commands...

```
Asset.joins(:project)
```
or
```
Asset.joins(:project).count
```

After running these commands we can see that we have 5 records of `Assets` associated with `Project`.
Now, let's run other commands:

```
Asset.joins(:project).where(id: nil)
```

We can see that none of the records has `id: nil`
**However**, when we execute the following command:

```
Asset.joins(:project).where(id: nil).delete_all
```

All records related to the the `Asset` model are deleted somehow! 😢 

```
Asset.count
```
Returns `0` now.

This is a serious bug of the composite_primary_keys gem and it is a problem related with the `subquery_for(key, select)` method that this gem overwrites from `ActiveRecord`. This problem is happening after this commit: https://github.com/kaefer-cit/composite_primary_keys/commit/d989a51def2519f89a0b4483120a340670bb783b

Before this commit, **it was not even possible to execute the `delete_all` after joining two tables**.

If someone can fix this, I'll be grateful.

Peace

Victorcorcos:
![screen shot 2018-04-17 at 15 47 17](https://user-images.githubusercontent.com/7637806/38890955-502525b2-4259-11e8-9e66-1b97bd31adab.png)

Victorcorcos:
I just removed the `composite_primary_keys` gem from my Gemfile and the `delete_all` command is working properly when a model is joined with another one before the deletion.

![screen shot 2018-04-17 at 18 03 45](https://user-images.githubusercontent.com/7637806/38896662-b92634f6-4269-11e8-8b59-fcf9e5670ed9.png)

Victorcorcos:
On the below picture you guys can check the different SQL queries and where the bug precisely takes place.

## With the gem (serious bug on `delete_all`)
![screen shot 2018-04-17 at 18 08 19](https://user-images.githubusercontent.com/7637806/38898494-8c650ee0-426b-11e8-91c3-f3165f78fc3f.png)

## Without the gem (everything is normal on `delete_all`)
[...]

PR #460, opened 2018-08-21
API change in AR 5.2.x

cpb:
[ActiveRecord::Associations::JoinDependency#aliases](https://github.com/rails/rails/commit/50036e673b3b6ff41476814b212b55fabac3a638#diff-06059df8d3dee3101718fb2c01151ad0R134) was made private.

A new method [#apply_column_aliases](https://github.com/rails/rails/commit/50036e673b3b6ff41476814b212b55fabac3a638#diff-06059df8d3dee3101718fb2c01151ad0R126) was added.

At first glance this new method appears to orchestrate a similar interaction as in the call replaced
below.

I submit this commit having checked the postgresql tests. I'll see what else I have available to test... Or hopefully the CI will take it!

Attempts to fix: https://github.com/composite-primary-keys/composite_primary_keys/issues/459

### DIFF
--- a/lib/composite_primary_keys/relation/finder_methods.rb
+++ b/lib/composite_primary_keys/relation/finder_methods.rb
@@ -18,7 +18,7 @@ module CompositePrimaryKeys
         end
 
         if block_given?
-          relation._select!(join_dependency.aliases.columns)
+          join_dependency.apply_column_aliases(relation)
           yield relation, join_dependency
         else
           relation


cpb:
Cool! Seems like just [TestDelete#test_delete_all_with_join](https://travis-ci.org/composite-primary-keys/composite_primary_keys/jobs/418527945#L965-L971) in `rake mysql:test` is failing!

kingdonb:
This PR unbroke my broken stuff that came with Rails 5.2.1 upgrade.  Thanks!  I'll take a look at the failing test and see if I can understand what else is wrong according to Travis.

kingdonb:
It looks like the failing test is associated with #444, and has nothing to do with this change.

Is this safe to merge for the next release of cpk?

cpb:
> It looks like the failing test is associated with #444, and has nothing to do with this change. (ref: [dd4d761](https://github.com/composite-primary-keys/composite_primary_keys/commit/dd4d761038fac66ba226de924f0b540fd12ca941))

Quickly scanning #444 it isn't clear that is resolved... The commit it references is broken on travis.

@kingdonb can you look at the diff between 5.1 and 5.2 in the AR part of the stack trace of the outstanding travis failure?

kingdonb:
I'll take a look a bit later, but as for this change, I think you'll find the same failure in an empty PR run against the master branch.  There was a fix that someone applied and it was reverted in #444, knowing it looks like they knew that would cause the test to fail (delete no longer possible on mysql through a join) but reverted the change anyway, because it was better to leave it broken than the alternative, which I guess was accidentally deleting too many things.

cfis:
Thanks!

PR #461, opened 2018-08-29
Fix sqlserver file loads on ActiveRecord 5.1

sigmike:
SQL Server monkey patch doesn't seem to work on ActiveRecord 5.1. This patch makes it work, but it probably breaks other adapters by requiring some sqlserver files. I'm not sure what's the right way to make it conditional.


cfis:
Looks ok to me...Thanks!

Issue #463, opened 2018-11-09
Error on counter_cache via relation with Rails 5.2.1


tomosm:
This is occurring with composite_primary_keys v11.0.3 and activerecord 5.2.1.

```
class Author < ActiveRecord::Base
  belongs_to :reading, counter_cache: :rating
end

class Reading < ActiveRecord::Base
  belongs_to :article
  belongs_to :user
end 
```

Then when I do this:

```
author = authors(:author1)
reading = readings(:santiago_first)
author.update(reading_id: reading.id)
```

I get the following error:

```
ActiveRecord::StatementInvalid: Mysql2::Error: You can't specify target table 'readings' for update in FROM clause: UPDATE `readings` SET `rating` = COALESCE(`rating`, 0) + 1 WHERE (`readings`.`id`) IN (SELECT `readings`.`id` FROM `readings` WHERE `readings`.`id` = 1)
```

Because `id` is a relation object from Rails 5.2.1., it works fine when I modify codes like this:

```
module ActiveRecord
  module CounterCache
    module ClassMethods
      def update_counters(id, counters)
...
        if primary_key.is_a?(Array)
...
        else
          if id.is_a?(ActiveRecord::Relation) && self == id.klass
            relation = id
          else
            relation = unscoped.where!(primary_key => id)
          end

          relation.update_all updates.join(", ")
        end
      end
    end
  end
end
```

cfis:
Yes, but that change then breaks CPK where id is a composite primary key. See if this change fixes the problem for you.

tomosm:
@cfis Thanks! It works. It would be nice if you could add some tests for counter_cash.

PR #464, opened 2018-11-09
Fix ActiveRecord::StatementInvalid: Mysql2::Error which is occurred b…

tomosm:
I try to fix this issue #463 . According to activerecord 5.2.1 implementation, `ActiveRecord::CounterCache::ClassMethods::update_counters` receives a relation object for `id` parameter. Please check `activerecord-5.2.1/lib/active_record/associations/builder/belongs_to.rb:L53`


cfis:
Fixed in an alternative manner - see ticket for comments.

Issue #466, opened 2018-12-10
Error when calling destroy_all on HABTM association


bradslaughter:
For example, if Assembly and Parts have the HABTM association from https://guides.rubyonrails.org/association_basics.html#the-has-and-belongs-to-many-association, the following code will raise this error on Rails 5.0 and 5.1:

```
> assembly = Assembly.create
> assembly.parts.create
> assembly.parts.destroy_all
ArgumentError: wrong number of arguments (1 for 0)
  from /Users/bslaught/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/arel-7.1.4/lib/arel/delete_manager.rb:3:in `initialize'
  from /Users/bslaught/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/composite_primary_keys-9.0.10/lib/composite_primary_keys/associations/has_many_through_association.rb:40:in `new'
  from /Users/bslaught/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/composite_primary_keys-9.0.10/lib/composite_primary_keys/associations/has_many_through_association.rb:40:in `delete_records'
  from /Users/bslaught/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/activerecord-5.0.7.1/lib/active_record/associations/collection_association.rb:535:in `remove_records'
  from /Users/bslaught/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/activerecord-5.0.7.1/lib/active_record/associations/collection_association.rb:528:in `block in delete_or_destroy'
  from /Users/bslaught/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/activerecord-5.0.7.1/lib/active_record/associations/collection_association.rb:205:in `block in transaction'
  from /Users/bslaught/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/activerecord-5.0.7.1/lib/active_record/connection_adapters/abstract/database_statements.rb:232:in `block in transaction'
  from /Users/bslaught/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/activerecord-5.0.7.1/lib/active_record/connection_adapters/abstract/transaction.rb:189:in `within_new_transaction'
[...]

PR #467, opened 2018-12-10
Fix destroy_all on HABTM in Rails 5

bradslaughter:
## Why?

`stmt = Arel::DeleteManager.new arel.engine` raises `ArgumentError: wrong number of arguments (1 for 0)` in Rails 5.0 and 5.1, and `undefined method 'engine' for #<Arel::SelectManager:0x007fbb79329ec0>` in Rails 5.2.

`Arel::DeleteManager#initialize` accepted an `engine` argument in arel 6 (https://www.rubydoc.info/gems/arel/6.0.0/Arel/DeleteManager), but not in arel 7 or higher (https://www.rubydoc.info/gems/arel/7.0.0/Arel/DeleteManager). `Arel::SelectManager#engine` was defined in arel 8 (https://www.rubydoc.info/gems/arel/8.0.0/Arel/SelectManager), but not arel 9 (https://www.rubydoc.info/gems/arel/9.0.0/Arel/SelectManager).

In addition, `ActiveRecord_AssociationRelation#bind_values` is no longer defined starting in Rails 4 (compare https://www.rubydoc.info/docs/rails/3.1.1/ActiveRecord/QueryMethods and https://www.rubydoc.info/docs/rails/4.0.0/ActiveRecord/QueryMethods), so `scope.bind_values` raises an undefined method error. 

## What?

The method raising these errors overrides Rails' `ActiveRecord::Associations::HasManyThroughAssociation#delete_records`. At the time this code was added to composite_primary_keys (May 26, 2014), we had the [same code as Rails' version of the method](https://github.com/rails/rails/blob/05a90c36c55fe0e43d60c2d0aa09d7a16ddd34ee/activerecord/lib/active_record/associations/has_many_through_association.rb), but we changed

```
scope.where! construct_join_attributes(*records)
```

to

```
if source_reflection.klass.composite?
  scope.where! cpk_join_through_predicate(*records)
else
  scope.where! construct_join_attributes(*records)
end
```

Given that, I see two approaches to fixing this:

1. Fix the bugs in the existing method. Since that method has [changed a lot in the past four years](https://github.com/rails/rails/commits/master/activerecord/lib/active_record/associations/has_many_through_association.rb), we may want to pull in other changes to fix more subtle issues and stay up-to-date with Rails' current functionality.

2. Override `construct_join_attributes` instead of overriding `delete_records`. This reduces the amount of Rails' code we're overriding, and therefore the risk of having to update this code again. The only other place this method is called is in `through_records_for` in the same file, but only when `record.composite?` is false, so behavior should be unchanged outside of `delete_records`.

I took the second approach in this PR, but if there is a good reason for overriding `delete_records` and the first approach is preferred, I have that change in https://github.com/composite-primary-keys/composite_primary_keys/compare/master...bradslaughter:destroy-all-habtm-alternative?expand=1 and can make a pull request with that branch instead.

## Which versions of composite_primary_keys are affected?

arel 7 is required by activerecord 5.0, arel 8 is required by activerecord 5.1, and arel 9 is required by activerecord 5.2, so we would want this change in composite_primary_keys major versions 9, 10, and 11.

## How can this fix be verified manually?

https://github.com/bradslaughter/cpk-habtm-destroy-all-example has an example Rails app where this bug can reproduced (steps are in the Github issue below). The bug is fixed with the changes on this branch.

## What are the relevant Github issues?

https://github.com/composite-primary-keys/composite_primary_keys/issues/466


cfis:
I am good with deleting the override for ActiveRecord::Associations::HasManyThroughAssociation#delete_records

And thanks for the test.

But instead of overriding construct_join_attributes, can you just fix it instead.  So create a new file for through_associations.rb in CPK and fix the original method. Make sure to comment each line with #CPK that has to change.

Thanks - Charlie

bradslaughter:
I agree `through_association.rb` is a better place for this. Added that change.

cfis:
Great - thanks for the fix!

PR #471, opened 2019-01-13
Fix #470: Updates due to changes in ActiveRecord

### LINKED ISSUE #470: 11.1 Broken on #find_each in Rails 5.2.2
When using `#find_each` on an model through an association, I get this error:

```
undefined method `act_on_order_or_limit_ignored' for #<MyModel::ActiveRecord_AssociationRelation:...>)
```

I'm not sure if it's only through an association or more general, but I don't have time to investigate at the moment, so I just wanted to report.

codeodor:


cfis:
Thanks!

PR #472, opened 2019-02-05
Add support for Rails 6

kajisha:
Rails 6 has not yet been released, but it will follow up changes of ActiveRecord 6.


nialbima:
Commenting to follow status here.

cfis:
Thanks for putting this together!  Will take a look and hopefully get it merged shortly.

Gemfile
```diff
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-gem 'activerecord', '~> 5.2.1'
+gem 'activerecord', '6.0.0.beta2'
```
javierjulio: @kajisha can you please update again as beta3 was recently released?
```suggestion
gem 'activerecord', '6.0.0.beta3'
```

kajisha: Thanks. I updated beta3 at https://github.com/composite-primary-keys/composite_primary_keys/pull/472/commits/0dc230a7b17cc2ed8041b7636220edaf7867e752

cfis:
Thanks for the pull request. This is a good first step for AR 6.0 support.

But with major upgrades, every overridden method needs to be checked against the new AR 6 version.  Also, there are 5 new test failures in associations and calculations.

So don't use this in production yet.

PR #473, opened 2019-02-05
Stay close to the original implementation of Association#run

f-mer:
The original implementation passes the key to the `convert_method` before fetching the owner. This commit restores the conversion behaviour.

Haven't figured out a case where this actually applies. Just hit this while trying to make preloading case insensitive by overriding this method :woman_shrugging: 

Not entirely sure what's wrong with the sqlite tests :thinking:


cfis:
Thanks!

PR #476, opened 2019-03-12
Describe unexpected behaviour when :id is in primary key

mberlanda:
We recently tried to integrate CPK on an existing project.
Some models behave like the PkUser model described in this test.

There a few things acting strangely and I tried to summarise  the issues with a minimal test

Update: renaming `:id` to `:user_id` fix the most of the unexpected behaviours => 6c2d19b8f6b8287c1d7f9852986003c5ebce8ff0


cfis:
Any chance you can use the existing test tables? I'm trying to reduce the number of tables, not increase them.

cfis:
Hopefully this has been fixed by b0413f5, so going to close this pull request.  Sorry it took so long to resolve.

PR #483, opened 2019-04-18
CI: Use 2.6.3, drop 2.4.6 (Rails 6 requires 2.5.0+)

### LINKED ISSUE #484: CI: Your version of PostgreSQL (90224) is too old.
Tests currently fail in Travis with an error message describing the problem:

> Your version of PostgreSQL (90224) is too old. Active Record supports PostgreSQL >= 9.3.

https://travis-ci.org/composite-primary-keys/composite_primary_keys/jobs/521793125#L539

olleolleolle:
This PR updates the CI configuration on the `master` branch:

- use Ruby 2.6.3 in the matrix
- use PostgreSQL 9.4 - Fixes #484 
- remove Ruby 2.4.6 from the matrix - Rails 6 requires at least Ruby 2.5.0, and we're fixed on that version
- remove the no-longer-used Travis setting `sudo: false`. See [more at the Travis blog](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration).

This makes the build 💚 green.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,12 +3,12 @@ bundler_args: --without "oracle sqlserver"
 
 before_script:
   - cp test/connections/databases.ci.yml test/connections/databases.yml
-  - rake mysql:build_database postgresql:build_database sqlite:build_database
+  - bundle exec rake mysql:build_database postgresql:build_database sqlite:build_database
 
 script:
-  - "rake postgresql:test"
-  - "rake sqlite:test"
-  - "rake mysql:test"
+  - "bundle exec rake postgresql:test"
+  - "bundle exec rake sqlite:test"
+  - "bundle exec rake mysql:test"
 
 env:
   global:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -11,7 +11,6 @@ script:
   - "rake mysql:test"
 
 rvm:
-  - 2.4.6
```
codeodor: Shouldn't we keep this for testing versions of CPK that are still valid but not on Rails 6 yet?

olleolleolle: Before I go any further with that, I'd like to focus on this section, in the very first part of the gem:

```ruby
unless defined?(ActiveRecord)
  require 'rubygems'
  gem 'activerecord', '6.0.0.beta3'
  require 'active_record'
end
```

codeodor: Sorry, I didn't realize `master` was already on 6+.  Never mind then!

cfis:
Thanks!

PR #485, opened 2019-05-01
Use latest Rails 6 rc1

ionut998:
Rails 6.rc1 fixed an issue about preloading associations [here](https://github.com/rails/rails/pull/35496)
I have ported over the changes to the Association class


cfis:
Thanks!

PR #489, opened 2019-07-02
Fix compatibility with AR 5.1.6+ re: optimistic locking

### LINKED ISSUE #447: active_record/locking/optimistic.rb doesn't have 'increment_lock' method
ActiveRecord 5.2 doesn't have increment_lock method, so maybe an error occurred in optimistic.rb.

```(ruby)
module ActiveRecord
  module Locking
    module Optimistic
      def _update_record(attribute_names = self.attribute_names)
        return super unless locking_enabled?
        return 0 if attribute_names.empty?

        begin
          lock_col = self.class.locking_column
          previous_lock_value = read_attribute_before_type_cast(lock_col)
          increment_lock
          attribute_names.push(lock_col)
          relation = self.class.unscoped

  ...
```

rlue:
Fixes #447.

### DIFF
--- a/lib/composite_primary_keys/persistence.rb
+++ b/lib/composite_primary_keys/persistence.rb
@@ -60,6 +60,14 @@ module ActiveRecord
           true
         end
       end
+
+      private
+
+      def increment_lock
+        lock_col = self.class.locking_column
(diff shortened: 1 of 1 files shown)


cfis:
Thanks!

PR #490, opened 2019-07-04
Reset id if part of it was generated during insertion

pawelniewie:
I found out that the gem doesn't reset id if part of it was already set before calling `create` this was causing trouble for me as I have a partitioned table by `tenant_id` but still use sequence to fill in a default for `id`.

This change resets `id` in that case. 

What do you think? Am I right doing it here?

### DIFF
--- a/lib/composite_primary_keys/persistence.rb
+++ b/lib/composite_primary_keys/persistence.rb
@@ -64,7 +64,7 @@ module ActiveRecord
       new_id = self.class._insert_record(attributes_values)
 
       # CPK
-      if self.composite? && self.id.compact.empty?
+      if self.composite? && self.id.compact.size != self.class.primary_keys.size
         self.id = new_id
       else
         self.id ||= new_id if self.class.primary_key


cfis:
That's a good question. Seems like overriding a set value might not be such a good plan.  Maybe just update nil values?  So if self.id.compact.count > 0, then fill in just the nil values?

pawelniewie:
@cfis what would be the rationale for updating only nils? I think getting all values from a new row is a simpler / cleaner approach but I probably miss some corner cases?

pawelniewie:
Also what's your strategy to handle all those different `ar_x.x.x` versions? I commited to a branch specific to the version I use. Should I backport it to other versions? What about master branch?

cfis:
I am wondering about the case when you intentionally set part of primary key to a specific value. We definitely would not want to overwrite that.  Maybe we could check using dirty attributes?

As for my strategy with different AR versions. Each version is generally quite different. So I never backpatch commits, its too hard/risky and possibly irrelevant.  If I add a test to an older version of AR, I will bring that change forward into newer branches.  But honestly that rarely happens since I 95% of my effort is on master branch.

pawelniewie:
> I am wondering about the case when you intentionally set part of primary key to a specific value. We definitely would not want to overwrite that. Maybe we could check using dirty attributes?

I'm doing that exactly in my case - I partitioned tables by tenant_id, so new records get one part set in the code before insert, the other part (sequence number) is automatically generated from a sequence on the database level.

The only scenario it wouldn't work as expected would be when you change the value set in the code on the database level too - for example if you have some trigger that does sth with the value passed from the code, but I assume that's rare and why would you do such a thing in the first place?

cfis:
Ok - to make sure we are on the same page.

Currently the gem only sets the id if all the values are empty (if self.composite? && self.id.compact.empty?)

I think what you want to do is set the ids even if they are partially filled in (self.id.compact.size != self.class.primary_keys.size).

But then doesn't that overwrite the values you previously set - which would be bad?

Instead, don't you want to fill in the id values that are nil (thus have not been set)?  If that's the case, I don't see how the patch accomplishes that.

PR #492, opened 2019-08-20
[WIP] Rails 6 compatible

ionut998:
Use version 6.0.0 of Rails

PR #496, opened 2019-09-02
Get all tests passing on MySQL and PostgreSQL on Rails 6.0.0

### LINKED ISSUE #495: ArgumentError: wrong number of arguments (given 1, expected 2) on 12.0.0.rc5
I'm not 100% sure this is due to CPK yet, but it seems like one of those strange errors in a Rails upgrade that I've seen due to CPK, so I just wanted to get something written down for later investigation while it's still a release candidate.

codeodor:
Fixes #495  and includes changes from #492


codeodor:
Can someone check the tests for the other supported databases?

codeodor:
CI is failing but appears due to a config issue: says it can't connect to MySQL through the socket.

codeodor:
Still getting some errors in my app's test suite. Not sure if they're related to CPK or not, yet. Will report back when I find out.

codeodor:
One error is related to a count query. Prior upgrade, it would `select count from (select distinct table.id1, table.id2 ... )...` but now it is dropping the `table` qualifier, which makes the `id1` and `id2` columns ambiguous (due to joining other tables which have those as columns too).

codeodor:
The major error in my tests is due to `includes`/`references` now coming after `joins` in the generated query. I don't know if I have a good way to tell in my app if that's due to something in CPK or something in Rails that changed.

codeodor:
I have not looked into the one related to the count yet, but the one related to `includes` I think is due to https://github.com/rails/rails/pull/36805

codeodor:
@cfis  I think this might be ready for a review, if you'd just like to double check me. 

The remaining issue in my app's test suite is that includes/references gets moved and further joins reference columns that aren't part of the query yet, but I am starting to think that is an issue with Rails itself, and I've been working around it by just calling `left_joins(:relationship_name)` and that seems to work.

cfis:
Looks ok to me.  Doesn't look like CPK overrides any of the methods changed in rails/rails#36805.

Would it be possible to add a test case for the includes/references issue?

codeodor:
I hope to get around to it. I still need to see if I can get it to work correctly in Rails without CPK. I thought I did, but when I ran what I thought was the same scenario in another Rails project, it actually just removed the includes/references altogether, so that's what makes me think it would be Rails at fault rather than CPK. Still need to do that research.

codeodor:
@cfis I confirmed the includes/references + join problem is an issue in Rails with this script, and reported to Rails: https://github.com/rails/rails/issues/37133

cfis:
Got it - thanks for reporting.

Issue #497, opened 2019-09-11
ORA-00920: invalid relational operator with CPK 12.0.0 against Oracle database


awaltman:
I cloned the CPK repository, configured the test connection for my Oracle database and ran the unit tests. There were many errors like the following:
```
TestCreate#test_has_many_ids_1:
ActiveRecord::StatementInvalid: OCIError: ORA-00920: invalid relational operator
    stmt.c:267:in oci8lib_260.so
    /home/awaltman/.rvm/gems/ruby-2.6.4/gems/ruby-oci8-2.2.7/lib/oci8/cursor.rb:137:in `exec'
    /home/awaltman/.rvm/gems/ruby-2.6.4/gems/activerecord-oracle_enhanced-adapter-6.0.0/lib/active_record/connection_adapters/oracle_enhanced/oci_connection.rb:146:in `exec_update'
    /home/awaltman/.rvm/gems/ruby-2.6.4/gems/activerecord-oracle_enhanced-adapter-6.0.0/lib/active_record/connection_adapters/oracle_enhanced/database_statements.rb:152:in `block in exec_update'
    /home/awaltman/.rvm/gems/ruby-2.6.4/gems/activerecord-6.0.0/lib/active_record/connection_adapters/abstract_adapter.rb:707:in `block (2 levels) in log'
    /home/awaltman/.rvm/rubies/ruby-2.6.4/lib/ruby/2.6.0/monitor.rb:235:in `mon_synchronize'
    /home/awaltman/.rvm/gems/ruby-2.6.4/gems/activerecord-6.0.0/lib/active_record/connection_adapters/abstract_adapter.rb:706:in `block in log'
    /home/awaltman/.rvm/gems/ruby-2.6.4/gems/activesupport-6.0.0/lib/active_support/notifications/instrumenter.rb:24:in `instrument'
    /home/awaltman/.rvm/gems/ruby-2.6.4/gems/activerecord-6.0.0/lib/active_record/connection_adapters/abstract_adapter.rb:697:in `log'
    /home/awaltman/.rvm/gems/ruby-2.6.4/gems/activerecord-oracle_enhanced-adapter-6.0.0/lib/active_record/connection_adapters/oracle_enhanced/dbms_output.rb:36:in `log'
[...]

PR #499, opened 2019-10-16
Fix for #497 (ORA-00920: invalid relational operator with CPK 12.0.0)

awaltman:
The name of the subquery alias on Oracle cannot start with the underscore character.
See: https://docs.oracle.com/en/database/oracle/oracle-database/12.2/sqlrf/Database-Object-Names-and-Qualifiers.html#GUID-75337742-67FD-4EC0-985F-741C93D918DA

This change also fixes an issue when deleting all records if the association
contains an order option.

### DIFF
--- a/lib/composite_primary_keys/relation.rb
+++ b/lib/composite_primary_keys/relation.rb
@@ -91,9 +91,6 @@ module ActiveRecord
         stmt.key = arel_attribute(primary_key)
         stmt.wheres = arel.constraints
       end
-      stmt.take(arel.limit)
-      stmt.offset(arel.offset)
-      stmt.order(*arel.orders)
 
       affected = @klass.connection.delete(stmt, "#{@klass} Destroy")
 
@@ -112,7 +109,7 @@ module ActiveRecord
 
       key_name = Array(key).map {|a_key| a_key.name }.join(',')
 
-      Arel::SelectManager.new(subselect.as("__active_record_temp")).project(Arel.sql(key_name))
+      Arel::SelectManager.new(subselect.as("active_record_temp")).project(Arel.sql(key_name))
     end
   end
 end
--- a/test/fixtures/restaurant.rb
+++ b/test/fixtures/restaurant.rb
@@ -1,6 +1,6 @@
 class Restaurant < ActiveRecord::Base
   self.primary_keys = :franchise_id, :store_id
-  has_and_belongs_to_many :suburbs, 
+  has_and_belongs_to_many :suburbs, -> { order("name ") },
     :foreign_key => [:franchise_id, :store_id],  
     :association_foreign_key => [:city_id, :suburb_id]
   has_and_belongs_to_many :products,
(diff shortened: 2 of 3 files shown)


lib/composite_primary_keys/relation.rb
```diff
@@ -91,9 +91,6 @@ def delete_all
         stmt.key = arel_attribute(primary_key)
         stmt.wheres = arel.constraints
       end
-      stmt.take(arel.limit)
```
codeodor: Removing this kind of scares me. Are there tests that show delete still works if there was a limit and and offset?

awaltman: Currently there are no tests for delete_all with an association with either limit or offset. I removed those lines based on comparing the versions of the methods from 11.2.0 (see: https://github.com/composite-primary-keys/composite_primary_keys/blob/55e9e51d5d0ca546184afb48939884a2207dcdc0/lib/composite_primary_keys/relation.rb#L53). It seems like the delete_all method in 12.0.0 was copied from update_all and I'm not sure if sort, limit, or offset make sense when attempting to delete all records. When testing deletes on an Oracle database it reports "ORA-00933: SQL command not properly ended" when attempting to use any of order by, limit or offset in the delete statement.

codeodor: We may need to look at how it looks in latest Rails. I have a vague memory of this code, but I could imagine if an update could take a limit and an offset, a delete might be expected to as well.  Not that we should write a bunch of special code to account for adapters that can't handle it, since it is kind of a special case, but I want to be sure if it was working in the other adapters, it would continue to work.

cfis: Check against the original ActiveRecord code of course to verify this change - that is what really matters. And please add a test, that should be fairly quick.

awaltman: Looking at ActiveRecord 6.0.0, the delete_all method in ActiveRecord::Relation also includes limit, offset and order bys:
```
      stmt = Arel::DeleteManager.new
      stmt.from(arel.join_sources.empty? ? table : arel.source)
      stmt.key = arel_attribute(primary_key)
      stmt.take(arel.limit)
      stmt.offset(arel.offset)
      stmt.order(*arel.orders)
      stmt.wheres = arel.constraints

      affected = @klass.connection.delete(stmt, "#{@klass} Destroy")
```
So I'm wondering if this works without errors on Oracle.

cfis: Good question - I'll let you verify that.  But if AR 6 has it, we can't remove it from CPK. Sorry.

cfis:
Just following up, where do we stand on this?

awaltman:
I'm going to go ahead and close this pull request and will create another request which includes just the fix for the alias name when I have more time. I apologize for the delay.

cfis:
No worries, thanks.

PR #502, opened 2019-11-26
Pk with default values

dannemanne:
If one of the composite keys have a default value, creating a new record results in the primary key attributes, not being assigned the database generated values from the incremental keys.

This is caused by ActiveRecords behaviour of pre-assigning default values on initialize which conflicts with this library's check of only assigning the primary keys if all of those attributes are nil. 

My tests showcase a scenario where one of the composite keys have an empty string as default value, but the other one is a normal incremental integer (could also be a uuid). This could be useful if you want one of the composite keys to act as the (primary) primary-key, but want to have the option of adding an alternative version for some of the records.

Since no column used as a primary key, can be NULL, it should be sufficient to check if any of the primary key attributes are nil, when determining if it should be assigned or not.

I have prepared tests for all the supported databases, but personally only tested this against the Postgres database.

### DIFF
--- a/lib/composite_primary_keys/persistence.rb
+++ b/lib/composite_primary_keys/persistence.rb
@@ -64,7 +64,7 @@ module ActiveRecord
       new_id = self.class._insert_record(attributes_values)
 
       # CPK
-      if self.composite? && self.id.compact.empty?
+      if self.composite? && self.id.any?(&:nil?)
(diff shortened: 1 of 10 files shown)


cfis:
Thanks! I changed the implementation just a bit (too just merge the specified id and new id).

PR #504, opened 2019-12-20
Only run the CPK code in update_all and delete_all when necessary

codeodor:
This resolves a performance issue on MySQL when running through the CPK version of the code.

Average test run on particular production data in my app showed a decrease from <unknown number because it was running for longer than 28 minutes so I stopped it> down to 928.0ms.

### DIFF
--- a/lib/composite_primary_keys/relation.rb
+++ b/lib/composite_primary_keys/relation.rb
@@ -30,6 +30,8 @@ module ActiveRecord
 
       # CPK
       if @klass.composite? && stmt.to_sql =~ /['"]#{primary_key.to_s}['"]/
+        stmt.table(table)
+
         arel_attributes = primary_keys.map do |key|
           arel_attribute(key)
         end.to_composite_keys
--- a/lib/composite_primary_keys/relation.rb
+++ b/lib/composite_primary_keys/relation.rb
@@ -25,10 +25,11 @@ module ActiveRecord
       end
 
       stmt = Arel::UpdateManager.new
-      # CPK
-      if @klass.composite?
-        stmt.table(table)
+      stmt.table(arel.join_sources.empty? ? table : arel.source)
+      stmt.key = arel_attribute(primary_key).name
 
+      # CPK
+      if @klass.composite? && stmt.to_sql =~ /['"]#{primary_key.to_s}['"]/
         arel_attributes = primary_keys.map do |key|
           arel_attribute(key)
         end.to_composite_keys
@@ -37,10 +38,9 @@ module ActiveRecord
 
         stmt.wheres = [arel_attributes.in(subselect)]
       else
-        stmt.table(arel.join_sources.empty? ? table : arel.source)
-        stmt.key = arel_attribute(primary_key)
         stmt.wheres = arel.constraints
       end
+
       stmt.take(arel.limit)
       stmt.offset(arel.offset)
       stmt.order(*arel.orders)
@@ -75,8 +75,10 @@ module ActiveRecord
       end
 
       stmt = Arel::DeleteManager.new
+      stmt.from(arel.join_sources.empty? ? table : arel.source)
+      stmt.key = arel_attribute(primary_key)
       # CPK
-      if @klass.composite?
+      if @klass.composite? && stmt.to_sql =~ /['"]#{primary_key.to_s}['"]/
         stmt.from(table)
 
         arel_attributes = primary_keys.map do |key|
@@ -87,8 +89,6 @@ module ActiveRecord
 
         stmt.wheres = [arel_attributes.in(subselect)]
       else
-        stmt.from(arel.join_sources.empty? ? table : arel.source)
-        stmt.key = arel_attribute(primary_key)
         stmt.wheres = arel.constraints
       end
       stmt.take(arel.limit)


codeodor:
This is working for me locally on Postgres 11.5. Gonna see if I can downgrade to see if that's the problem on the CI, which uses 9.6.

codeodor:
Ok, fixed. @cfis  what do you think?

cfis:
Seems pretty ugly :(  

And I'm not seeing why its necessary. Doesn't the .composite? call work?

codeodor:
@cfis  The problem is when it's running through this code on my production system it's taking over 30 minutes to run. With this fix, it runs in a few milliseconds. 

I guess mysql does not like it when you have code like `where (id1, id2, id3) = (1, 2, 3)`? 

In particular, it's running through this code even when it is not necessary for the task. I wasn't able to ascertain why it was running when not necessary, but basically, when I add this condition in I get rid of a huge performance issue and everything else still works.

cfis:
Right, I understand the issue. I don't understand why it would happen though, since the cpk check should be fast and avoid it.  Suppose you don't have an easily reproducible test case?

codeodor:
No, I think it requires more data that I could include in a test case. 

But basically what this is doing is forcing an inner query when one is not needed.

For example, suppose I have this code where I want to update or delete all users not on the primary key, but some other criteria, and `User` has a composite key:

`User.where('updated_at < ?', Time.now - 1.year).delete_all`

That generates a query like this:

`delete from users where (users.cpk1, users.cpk2, users.cpk3) in (select cpk1, cpk2, cpk3 from users where updated_at < '2019-02-25')`

But there is no reason for that. It should just generate a query like this:

`delete from users where updated_at < '2019-02-25'`

MySQL is already pretty notoriously bad in some cases with sub selects like this, and then when it uses a comparison of several columns like this, it gets even worse, is my thinking. It probably gets worse in my case because rather than 2 columns, the composite key is like probably 5-7 fields. 

So basically, the proposed solution is to use the CPK code here, not _just_ if the class is composite, but also _only_ when it's needed -- i.e., when the relation on which we're trying to update all or delete all actually was already looking up things by the primary key.

codeodor:
No, I think it requires more data that I could include in a test case. 

But basically what this is doing is forcing an inner query when one is not needed.

For example, suppose I have this code where I want to update or delete all users not on the primary key, but some other criteria, and `User` has a composite key:

`User.where('updated_at < ?', Time.now - 1.year).delete_all`

That generates a query like this:

`delete from users where (users.cpk1, users.cpk2, users.cpk3) in (select cpk1, cpk2, cpk3 from users where updated_at < '2019-02-25')`

But there is no reason for that. It should just generate a query like this:

`delete from users where updated_at < '2019-02-25'`

So, yes, the `composite?` call works in that the class is indeed composite. The problem is it's still not necessary to run through that code. 

MySQL is already pretty notoriously bad in some cases with sub selects like this, and then when it uses a comparison of several columns like this, it gets even worse, is my thinking. It probably gets worse in my case because rather than 2 columns, the composite key is like probably 5-7 fields. 

So basically, the proposed solution is to use the CPK code here, not _just_ if the class is composite, but also _only_ when it's needed -- i.e., when the relation on which we're trying to update all or delete all actually was already looking up things by the primary key.

cfis:
Ah that makes sense.  Never liked that subquery anyway, wonder if its possible to get rid of it. I vaguely remember it wouldn't work otherwise on some database, maybe it was sqlite?

codeodor:
Sorry for my late reply. Probably SQLite and others. I actually had never seen it before this, although IMO it makes sense and is a nice way to easily represent ANDs like `in` does for ORs.

As for getting rid of it, I think it is possible, but we have to convert it to a bunch of ANDs. 

I am surprisingly swamped right now but I did want to say I've been looking at all these PRs and you are a saint as far as reviewing them all. Thank you for your efforts on CPK.

cfis:
Can you see if what I just checked is still ok? It uses WHERE () IN (SELECT ..) still like you implemented, but one less level of nesting. I played around with using EXISTS, but that didn't really work that well across all databases.

Anyway, going to close this. Reopen if needed.

codeodor:
@cfis I will test it out, leaving this here so I'll remember: https://github.com/composite-primary-keys/composite_primary_keys/commit/1e238902e5a5df97eb344926869d5af6ccda3787

codeodor:
Ok, so I finally got around to testing this and it seems 1e23890 broke the build for mysql. I am getting a lot of errors in the tests in my work on making it work for ActiveRecord 6.1 that seem to be related to `You can't specify target table '...' for update in FROM clause` where it is generating code like:

```
UPDATE `tariffs` SET `tariffs`.`amount` = COALESCE(`tariffs`.`amount`, 0) - 1 WHERE (`tariffs`.`tariff_id`, `tariffs`.`start_date`) IN 
(SELECT `tariffs`.`tariff_id`, `tariffs`.`start_date` FROM `tariffs` WHERE `tariffs`.`tariff_id` = 1 AND `tariffs`.`start_date` = '2020-11-26')
```

However, the commit just before that one, af9896014b8b1d7d01542ca6e17b56c0e171b1e8, the tests pass.

Issue #505, opened 2019-12-27
Critical bug: comma is not allowed as a value of attribute which is a part of composite key


timanovsky:
No idea how this trivial case have not been caught before. Tested with rails 4.2 and gem 8.1.0 - 8.1.7, but according to source code other versions should behave the same.

**Table**
```
  Table "public.test_models"
 Column |  Type   | Modifiers 
--------+---------+-----------
 c1     | integer | not null
 c2     | text    | not null
Indexes:
    "test_models_pkey" PRIMARY KEY, btree (c1, c2)
```

**Model** 
```ruby
class TestModel < ApplicationRecord
  self.primary_keys = :c1, :c2
end
```

**Repro**
```
> m = TestModel.create!(c1: 2, c2: "bbb,cccc")
=> #<TestModel:0x000000001a662a78 c1: 2, c2: "bbb,cccc">
> m.reload
NoMethodError: undefined method `length' for 2:Fixnum
from /home/timanovsky/.rvm/gems/ruby-2.3.8/gems/composite_primary_keys-8.1.7/lib/composite_primary_keys/relation/finder_methods.rb:140:in `block in find_some'
```

This is because this code for some reason parses attribute values recursively and splits strings : https://github.com/composite-primary-keys/composite_primary_keys/blob/master/lib/composite_primary_keys/composite_arrays.rb#L11

So `[2, "bbb,cccc"]` becomes `[2, ["bbb", "cccc"]]`

cfis:
Interesting, guess no one has run into that before. Wonder if its true on later versions too.  I commented on the patch...

cfis:
These changes were merged in and also ported to master branch.

sitnikovme:
@cfis any plans for making this critical fix available for version 11 of the gem for rails 5?

sitnikovme:
@cfis could you please take a look and merge https://github.com/composite-primary-keys/composite_primary_keys/pull/510 ?

cfis:
Yes - done.

PR #506, opened 2019-12-30
Fix issue #505 pk with comma

Slike9:
The commit fixes an issue when a CPK field contains commas.
Issue: https://github.com/composite-primary-keys/composite_primary_keys/issues/505

CPK is represented by Array or String. Fields in CPK as String are separated by comma. If a CPK field contains a comma, string representation of such CPK is parsed incorrectly. To address the issue,
commas in fields of CPK represented by String are escaped.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,12 @@ rvm:
   - 2.0.0
   - 2.1
   - 2.2
+  - 2.3
 
 env:
-  - CPK_LOGFILE=log/test.log
+  global:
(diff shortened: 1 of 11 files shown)


Slike9:
> Can this be slimmed down to just fix the problem? For example, changing blocks from the style of do .. end to {} is extra codes that hides the intent of the patch but does nothing to actually fix it.

@cfis Fixed

cfis:
Cool - thanks!

PR #507, opened 2020-01-07
Remove extra argument from sql_for_insert

tohae:
I fixed method arguments for Rails 6 as well as other adapters.

FYI: https://github.com/composite-primary-keys/composite_primary_keys/commit/9a51742b3b3c6c967d3ad2408d80c699dff69be5#diff-52b23807f8898e851aad8bf740baca75


cfis:
Thanks!  Happy to take any and all sql server patches, I don't test it very often.

PR #512, opened 2020-03-12
Fix overriding AbstractReflection for activerecord 5.2.4

Slike9:
In https://github.com/rails/rails/commit/54de9b1e4f5823bf22001be60efcb996ce6d260b (v5.2.4)
`ActiveRecord::Reflection::AbstractReflection#build_join_constraint`
method was removed.

Fixes incorrect SQL condition for joining by CPK.

### DIFF
--- a/composite_primary_keys.gemspec
+++ b/composite_primary_keys.gemspec
@@ -27,7 +27,7 @@ Gem::Specification.new do |s|
   # Dependencies
   s.required_ruby_version = '>= 2.2.2'
 
-  s.add_dependency('activerecord', '~> 5.2.1')
+  s.add_dependency('activerecord', '~> 5.2.4')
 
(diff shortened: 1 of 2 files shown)


composite_primary_keys.gemspec
```diff
@@ -27,7 +27,7 @@ Gem::Specification.new do |s|
   # Dependencies
   s.required_ruby_version = '>= 2.2.2'
 
-  s.add_dependency('activerecord', '~> 5.2.1')
+  s.add_dependency('activerecord', '~> 5.2.4')
```
Slike9: @cfis 
Suspect it makes sense to use
```
s.add_dependency('activerecord', '~> 5.2.1', '< 5.2.4')
```
for current version of the gem.

cfis:
Thanks!

PR #513, opened 2020-03-20
Active record 5.2.4 compatability

eanders:
This is based on https://github.com/composite-primary-keys/composite_primary_keys/pull/512.  Working to get the tests for MySQL and SQL Lite passing.

From #512:
In rails/rails@54de9b1 (v5.2.4)
ActiveRecord::Reflection::AbstractReflection#build_join_constraint
method was removed.

Fixes incorrect SQL condition for joining by CPK.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,8 @@ before_script:
 
 script:
   - "rake postgresql:test"
+  - "rake sqlite:test"
+  - "rake mysql:test"
 
 rvm:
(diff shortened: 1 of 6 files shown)


cfis:
Since #512 is applied I assume this is no longer necessary?

As for the schema updates, make those a separate pull request. That would make it easier to apply to master branch too.

Thanks - Charlie

PR #515, opened 2020-04-01
Fixes for Travis tests - AR 5.2.x

eanders:
Charlie, let me know if you'd prefer the schema updates separated from the test fixes.  The tests won't pass without those two changes, and I'm not sure how necessary the schema changes are without the tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,11 @@ language: ruby
 sudo: false
 bundler_args: --without "db2 oracle sqlserver"
 
+services:
+   - mysql
+   - postgresql
+   - sqlite3
+
 before_script:
   - cp test/connections/databases.ci.yml test/connections/databases.yml
   - rake mysql:build_database postgresql:build_database sqlite:build_database
@@ -12,10 +17,10 @@ script:
   - "rake mysql:test"
 
 rvm:
-  - 2.4.10
+  - 2.4.9
   - 2.5.8
-  - 2.6.6
-  - 2.7.1
+  - 2.6.5
+  - 2.7.0
 
 env:
   - CPK_LOGFILE=log/test.log
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,6 @@ language: ruby
 sudo: false
 bundler_args: --without "db2 oracle sqlserver"
 
-services:
-   - mysql
-   - postgresql
-   - sqlite3
-
 before_script:
   - cp test/connections/databases.ci.yml test/connections/databases.yml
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -2,6 +2,11 @@ language: ruby
 sudo: false
 bundler_args: --without "db2 oracle sqlserver"
 
+services:
```
Slike9, line 5: The file already has `services` section bellow..

test/test_create.rb
```diff
@@ -50,9 +50,8 @@ def test_create_no_id
 
   def test_create_generated_keys
     # Not all databases support columns with multiple identity fields
-    if defined?(ActiveRecord::ConnectionAdapters::PostgreSQL) ||
-       defined?(ActiveRecord::ConnectionAdapters::SQLite3)
-
+    postgres = defined?(ActiveRecord::ConnectionAdapters::PostgreSQL) && Suburb.connection.class == ActiveRecord::ConnectionAdapters::PostgreSQL
+    if postgres
```
Slike9, line 54: What about using `skip`?

test/fixtures/db_definitions/mysql.sql
```diff
@@ -52,7 +52,7 @@ create table product_tariffs (
 );
 
 create table suburbs (
-    city_id int not null,
+    city_id int not null auto_increment,
```
Slike9, line 55: Suspect we don't need this change as we are skipping the test.

eanders:
@Slike9 Thanks for the feedback.  I've adjusted and it appears all tests are still passing.

.travis.yml
```diff
@@ -12,14 +12,15 @@ script:
   - "rake mysql:test"
 
 rvm:
-  - 2.4.10
+  - 2.4.9
   - 2.5.8
-  - 2.6.6
-  - 2.7.1
+  - 2.6.5
+  - 2.7.0
```
codeodor, line 18: Why are these being decremented?

test/fixtures/db_definitions/mysql.sql
eanders: This is used in other tests which will fail without it.

.travis.yml
eanders: At the time of writing, travis was reporting not having access to the newer rubies.  That seems to have been fixed.  I've pushed an updated travis.yml.

cfis:
Yes, let's split this up. One patch for Travis. As for the schema changes, I'm not understanding the point of the Suburb check?

Going to close this pull request - let's separate these out.

PR #519, opened 2020-04-27
Closes #518 Fix issue with ORM default false values

### LINKED ISSUE #518: Default false value not setting correctly on create
I am experiencing an issue where I run `Model.create` with a column in the composite index that is defaulted to `false`.  The object returned has this field set to `nil`. The record is correctly created, but the object instantiated is not.  

This is occurring in v11.3.1

I have tracked the change to this:
 
https://github.com/composite-primary-keys/composite_primary_keys/commit/e7dab1432747c0ade3ca13c36d6f7ac8069418e2

I was wondering why this change was made to merge the changes together.  I believe this is in error since `(false || nil)` would return `nil`. 

Ruby: 2.6.6
Rails: 5.2.4.2
CPK: 11.3.1
Mysql: 5.7

andrewyoo:
This fixes the issue when a field in the composite primary key has a default value of `false`.  See #518   

@cfis I kind of googled and updated the different flavors of DBs, but only really tested on mysql.  Perhaps you have some notes i can update this PR with?

### DIFF
--- a/lib/composite_primary_keys/persistence.rb
+++ b/lib/composite_primary_keys/persistence.rb
@@ -66,7 +66,7 @@ module ActiveRecord
       # CPK
       if self.composite?
         # Merge together the specified id with the new id (specified id gets precedence)
-        self.id = self.id.zip(Array(new_id)).map {|id1, id2| (id1 || id2)}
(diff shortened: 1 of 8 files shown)


cfis:
Thanks!

PR #523, opened 2020-06-10
Add test for #491

### LINKED ISSUE #491: Polymorphic association does not support to compute class
In Rails 5.2 and up, you cannot call `klass` on a through reflection; it will return the mentioned error. [CPK does this anyway](https://github.com/composite-primary-keys/composite_primary_keys/blob/55e9e51d5d0ca546184afb48939884a2207dcdc0/lib/composite_primary_keys/associations/through_association.rb#L8). When removing records from a hm through association, errors will occur when records are removed on removing records from the association.

As an example, consider the following application records:

```ruby
class Property < ApplicationRecord; end
class Facility < Property; end

class Todo < ApplicationRecord
  belongs_to :property, polymorphic: true, required: true
end

class User < ApplicationRecord
  has_many :todos
  has_many :facilities, through: :todos, source: :property, source_type: 'Facility'
end
```

When clearing facilities of a user that has some facilities, for example using `user.facility_ids = []`, the error will occur with the following trace:

```
activerecord (5.2.3) lib/active_record/reflection.rb:420:in `compute_class'
activerecord (5.2.3) lib/active_record/reflection.rb:379:in `klass'
composite_primary_keys (11.2.0) lib/composite_primary_keys/associations/through_association.rb:8:in `construct_join_attributes'
activerecord (5.2.3) lib/active_record/associations/has_many_through_association.rb:135:in `delete_records'
activerecord (5.2.3) lib/active_record/associations/collection_association.rb:398:in `remove_records'
activerecord (5.2.3) lib/active_record/associations/has_many_through_association.rb:108:in `remove_records'
activerecord (5.2.3) lib/active_record/associations/collection_association.rb:391:in `block in delete_or_destroy'
activerecord (5.2.3) lib/active_record/associations/collection_association.rb:135:in `block in transaction'
activerecord (5.2.3) lib/active_record/connection_adapters/abstract/database_statements.rb:265:in `transaction'
activerecord (5.2.3) lib/active_record/transactions.rb:212:in `transaction'
activerecord (5.2.3) lib/active_record/associations/collection_association.rb:134:in `transaction'
activerecord (5.2.3) lib/active_record/associations/collection_association.rb:391:in `delete_or_destroy'
activerecord (5.2.3) lib/active_record/associations/collection_association.rb:189:in `delete'
activerecord (5.2.3) lib/active_record/associations/collection_association.rb:413:in `replace_records'
activerecord (5.2.3) lib/active_record/associations/collection_association.rb:251:in `block in replace'
activerecord (5.2.3) lib/active_record/associations/collection_association.rb:135:in `block in transaction'
activerecord (5.2.3) lib/active_record/connection_adapters/abstract/database_statements.rb:265:in `transaction'
activerecord (5.2.3) lib/active_record/transactions.rb:212:in `transaction'
activerecord (5.2.3) lib/active_record/associations/collection_association.rb:134:in `transaction'
activerecord (5.2.3) lib/active_record/associations/collection_association.rb:251:in `replace'
composite_primary_keys (11.2.0) lib/composite_primary_keys/associations/collection_association.rb:26:in `ids_writer'
activerecord (5.2.3) lib/active_record/associations/builder/collection_association.rb:65:in `facility_ids='
```

I'm not sure if there is a good solution for this. I'm using the following patch to fix this now (of course I know in this case that CPK does not apply):

```ruby
require 'composite_primary_keys/associations/through_association'

module ActiveRecord
  module Associations
    module ThroughAssociation
      alias cpk_construct_join_attributes construct_join_attributes

      def construct_join_attributes(*records)
        if source_reflection.polymorphic?
          original_construct_join_attributes(*records)
        else
          cpk_construct_join_attributes(*records)
        end
      end
    end
  end
end
```

f-mer:
This is a first step to fix #491.

Replacing `if source_reflection.klass.composite?` with `if !source_reflection.polymorphic? && source_reflection.klass.composite?` as suggested by @moiristo makes the tests pass but I'm feeling like this could causes different problems.

@cfis you've got an idea here? :see_no_evil:

### DIFF
--- a/test/fixtures/hack.rb
+++ b/test/fixtures/hack.rb
@@ -1,5 +1,6 @@
 class Hack < ActiveRecord::Base
   has_many :comments, :as => :person
-  
+  has_many :users, :through => :comments, :source => :person, :source_type => "User"
+
   has_one :first_comment, :as => :person, :class_name => "Comment"
 end
--- a/test/test_associations.rb
+++ b/test/test_associations.rb
@@ -174,6 +174,12 @@ class TestAssociations < ActiveSupport::TestCase
     assert_equal(3, tarrifs_length)
   end
 
+  def test_has_many_through_assign_ids
+    hack = Hack.new(name: "foo", users: User.all)
+    hack.user_ids = [""]
+    assert_empty(hack.user_ids)
+  end
+
   def test_new_style_includes_with_conditions
     product_tariff = ProductTariff.includes(:tariff).where('tariffs.amount < 5').references(:tariffs).first
     assert_equal(0, product_tariff.tariff.amount)


cfis:
Thanks for the test. So if this gets merged in, the test will fail (which is ok, just verifying).

Can you add a little bit of tests data so this through association has a bit of data on the other side?

As for a fix - good question. I've never actually used polymorphic associations, so not sure.

cfis:
Thanks for the inspiration. I decided to remove the hack table, so did not merge this test. Instead, I updated the data model a bit to link comments to articles and then I could implement a similar test.

Thanks for the help on this!

PR #524, opened 2020-06-12
Fix subquery_for for non-cpk

kellydaniels:
Within the subquery_for method, 'key' is an array of structs (Arel::Attributes::Attribute) for cpk models and a single struct for non-cpk models. It seems that the code is calling Array(key) assuming that in the case of a single struct you will have an array with one element, which is the struct. Instead, calling [struct.to_a](https://apidock.com/ruby/Struct/to_a) actually takes the struct and turns it into an array, where each field of the array is a field from the struct. Example:
Screenshot from 2020-06-12 13-38-49

To achieve the intended result, we can instead call [key].flatten. Example:
Screenshot from 2020-06-12 13-39-11

For reference, here are the same two examples, but with a cpk:
Screenshot from 2020-06-12 13-33-34
Screenshot from 2020-06-12 13-34-06

### DIFF
--- a/lib/composite_primary_keys/connection_adapters/abstract_mysql_adapter.rb
+++ b/lib/composite_primary_keys/connection_adapters/abstract_mysql_adapter.rb
@@ -11,7 +11,7 @@ module ActiveRecord
 
         # CPK
         #key_name = quote_column_name(key.name)
-        key_name = [key].flatten.map {|a_key| quote_column_name(a_key.name)}.join(',')
+        key_name = Array.wrap(key).map {|a_key| quote_column_name(a_key.name)}.join(',')
 
         Arel::SelectManager.new(subselect.as("__active_record_temp")).project(Arel.sql(key_name))
       end
--- a/lib/composite_primary_keys/connection_adapters/abstract_mysql_adapter.rb
+++ b/lib/composite_primary_keys/connection_adapters/abstract_mysql_adapter.rb
@@ -11,7 +11,7 @@ module ActiveRecord
 
         # CPK
         #key_name = quote_column_name(key.name)
-        key_name = Array(key).map {|a_key| quote_column_name(a_key.name)}.join(',')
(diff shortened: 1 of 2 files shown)


codeodor:
@kellydaniels Thanks! Does this fix an existing issue ... or if it fixes an error, could you possibly include a test case that goes from failing to fixed?

kellydaniels:
@codeodor Thanks for reviewing! Sorry, forgot to commit that. I've updated and added a test.

codeodor:
What do you think @cfis? Looks good to me. Failure appears unrelated.

However, I wonder if the failure is related to another PR recently? I don't remember if it got merged but I thought there was some discussion about autogenerating IDs. Sorry, been tough to keep up with, but I'm trying to do at least prelim work like asking for tests :)

codeodor:
I also wonder if this would end up fixing the [performance issue I had with CPK](https://github.com/composite-primary-keys/composite_primary_keys/pull/504) on a different branch.

cfis:
I don't like this change:

Array(key)

To:

[key].flatten

The whole point of the Array method is if key is an array than it does nothing otherwise it create a one-element array (with key as the first method). This is much better (in my view) then having to always create an array and then have to flatten it. Ruby gives us a method just for that (Array), so let's keep using it.

So this seems like a step backwards to me. I do like the added test though, happy to keep that.

kellydaniels:
The issue is that Array() isn't doing what you said in this case. Array() calls to_a on the arg. In this case the arg is a struct, turning it into a multi element array, with each element of the struct being an element in the array, not converting it to a single element array. You can see this if you run my test with and without the change. 

I understand that [].flatten isn't a pretty change. At first I thought that Array.wrap() was going to have the same issue, but looking now I see that it actually works properly, so I can change this to a Array.wrap. Thoughts?

kellydaniels:
@cfis I've changed it to wrap(). Let me know what you think.

cfis:
I see - that makes sense.

PR #527, opened 2020-09-14
Implement raw access for attributes

### LINKED ISSUE #526: AttributeMethods has_attribute? override conflicts with other gems
Problem:
    `composite_primary_keys`'s override of ActiveRecord::AttributeMethods.has_attribute? uses AttributeMethods.attributes instead of @attributes as the original method it overrides does. This conflicts with gems which override AttributeMethods.attributes, creating a situation which will cause a SystemStackError.

Fix:
  In the gem's override of has_attribute?, convert the `attributes` call to a direct access to `@attributes`.

ndguarino:
* Prevents infinite loops with gems which modify the 'attributes' method.
   * Fixes #526


cfis:
Thanks!

PR #528, opened 2020-09-30
fix abstract adapter (mysql) create with auto-increment

### LINKED ISSUE #518: Default false value not setting correctly on create

ryantm:
fixes #518

With rails 6.0.3.3, cpk 12.0.2, and MariaDB 10.4.14, `exec_insert` with does
not populate value.rows, so we need to use last_inserted_id to get the
id.

Also, self.id is nil, so we need to guard against that case in
Persistence.

### DIFF
--- a/lib/composite_primary_keys/connection_adapters/abstract/database_statements.rb
+++ b/lib/composite_primary_keys/connection_adapters/abstract/database_statements.rb
@@ -6,7 +6,7 @@ module ActiveRecord
         value = exec_insert(sql, name, binds, pk, sequence_name)
 
         # CPK
-        if value && pk.is_a?(Array)
+        if !value&.rows&.empty? && pk.is_a?(Array)
           id_value || value.rows.first
         else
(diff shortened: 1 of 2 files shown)


cfis:
Is that a change in exec_insert? If so, shouldn't that be fixed instead?

cfis:
What would be helpful is a test case....

PR #530, opened 2020-11-14
Fix references to RecordNotFound

Slike9:
The change fixes error on referencing `RecordNotFound` inside `CompositePrimaryKeys::ActiveRecord::FinderMethods`:
 ```
 NameError: uninitialized constant CompositePrimaryKeys::ActiveRecord::FinderMethods::RecordNotFound   
 ```


cfis:
Cool - thanks! Looks like master branch might have the same issue, will take a look at that.

PR #532, opened 2020-12-03
Stay compatible with ruby 2.6 and below

ta1kt0me:
Calling a private method with `self` receiver is supported from ruby 2.7.

In ruby 2.6 it resolves a below error,

```
NoMethodError: private method `column_for' called for #<ActiveRecord::ConnectionAdapters::Mysql2Adapter:0x000055bd449c00e8>
```

refs
- https://github.com/ruby/ruby/pull/2474
- https://bugs.ruby-lang.org/issues/11297
- https://bugs.ruby-lang.org/issues/16123


cfis:
Thanks!

Issue #537, opened 2020-12-31
Switch to Github Actions For Tests


cfis:
Since tarvis.org is going away, we need to switch to travis.com or Github actions. I tried moving to Travis.com last month but failed. Seems like Github actions is worth a try, I see other projects moving to it.

PR #540, opened 2021-01-22
Switch CI from Travis CI to GitHub Actions

ta1kt0me:
When merging a library update such as renovate into our product, I wanted to make sure that the CI of the library passed, so I worked on #537 .
Currently, Travis CI results on this project continue to fail, so I want to improve the situation where it is difficult to confirm that library changes are safe.

Since I think that using multiple different DBs at the same time in Rails application is not a realistic use case, so it defines a workflow for each DB.


.github/workflows/mysql.yml
```diff
@@ -0,0 +1,45 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{matrix.ruby}}
+          bundler-cache: true
+      - name: Setup database config
+        run: cp test/connections/databases.ci.yml test/connections/databases.yml
+      - name: Prepare databases
+        run: bundle exec rake mysql:rebuild_database
```
ta1kt0me, line 41: It uses the environment variable `MYSQL_DATABASE` to authorize the db user who connects to the mysql service, but uses `mysql:rebuild_database` instead of `mysql:build_database` to also create a database at the same time.

cfis:
This looks awesome. So once this is merged, what would be the next steps? Does it just work with Github actions or do we need to go setup something with Github?

ta1kt0me:
I think it enables GitHub Actions in the origin repository as well after merge. And it's good to limit permissions. Could you try below steps?

1. Go to Settings > Actions
2. Check `Allow select actions`, then select both `Allow actions created by GitHub` and `Allow Marketplace actions by verified creators` in`Actions permissions` section

cfis:
Awesome. Merged and actions updated as you specified. Actions are now running!

PR #550, opened 2021-03-22
Use original implementation to execute COUNT without subquery.

hrysd:
When operation like followings, CPK generates subquery always.
And I don't think `ActiveRecord::Calculations#execute_simple_calculation` needs CPK specific implementation.
So I stop to overwrite this.
And I also don't think overwriting this need on master too. If this PR can merge.  I'll create same PR to master

```
User.count(:id)
```

```
SELECT COUNT(count_column) FROM (SELECT `users`.`id` AS count_column FROM `users`) subquery_for_count
```

### DIFF
--- a/lib/composite_primary_keys/relation/calculations.rb
+++ b/lib/composite_primary_keys/relation/calculations.rb
@@ -14,46 +14,6 @@ module CompositePrimaryKeys
         end
       end
 
-      def execute_simple_calculation(operation, column_name, distinct) #:nodoc:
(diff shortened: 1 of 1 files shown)


hrysd:
As far as see test result, I think `ActiveRecord::Calculation::aggregate_column`  is also unnecessary. but executed SQL could not resolve dup column. I'll investigate this and recreate PR.

cfis:
Sounds good.

PR #560, opened 2021-08-11
add scoped association test

ryantm:
I'm having a MariaDB SQL generation problem where my scoped association with composite primary keys gets generated with SQL like:

```
FROM `table1` INNER JOIN `table2` ON  AND
```

ON and AND are not supposed to be next to one another.

I'm trying to make a test case that uncovers this. I haven't figured out how to run the tests locally, so I'm making this draft PR to try to use GitHub actions to run my test.

### DIFF
--- a/lib/composite_primary_keys/associations/join_dependency.rb
+++ b/lib/composite_primary_keys/associations/join_dependency.rb
@@ -1,6 +1,25 @@
 module ActiveRecord
   module Associations
     class JoinDependency
+
+      class JoinAssociation < JoinPart # :nodoc:
+        private
+          def append_constraints(join, constraints)
+            if join.is_a?(Arel::Nodes::StringJoin)
+              join_string = Arel::Nodes::And.new(constraints.unshift join.left)
+              join.left = Arel.sql(base_klass.connection.visitor.compile(join_string))
+            else
+              right = join.right
+              # CPK
+              if right.expr.children.empty?
+                right.expr = Arel::Nodes::And.new(constraints)
+              else
+                right.expr = Arel::Nodes::And.new(constraints.unshift right.expr)
+              end
+            end
+          end
+      end
+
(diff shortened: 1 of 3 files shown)


ryantm:
Yep, this triggers the problem. I ran the actions on my fork:

https://github.com/ryantm/composite_primary_keys/runs/3303330479?check_suite_focus=true#step:7:598

and it generated the syntax error:

```
D, [2021-08-11T16:55:53.656361 #2584] DEBUG -- :   SQL (0.5ms)  SELECT `memberships`.`user_id` AS t0_r0, `memberships`.`group_id` AS t0_r1, `membership_statuses`.`id` AS t1_r0, `membership_statuses`.`user_id` AS t1_r1, `membership_statuses`.`group_id` AS t1_r2, `membership_statuses`.`status` AS t1_r3 FROM `memberships` LEFT OUTER JOIN `membership_statuses` ON  AND `membership_statuses`.`user_id` = `memberships`.`user_id` AND `membership_statuses`.`group_id` = `memberships`.`group_id` AND `membership_status`.`status` = 'Active'
```

ryantm:
The syntax is also bad for postgresql and sqlite.

codeodor:
Thank you @ryantm. I am swamped so cannot afford to look at a fix in depth at the moment, but wanted to suggest a workaround of using something like `where('membership_statuses.active = ?', true)` for the moment -- would that work?

ryantm:
Thanks for the quick reply and all you've done for this gem. That workaround does not fix the problem. https://github.com/ryantm/composite_primary_keys/pull/2/checks?check_run_id=3303749122#step:7:759

I'm still trying to work on a fix myself, but my AREL debugging skills need upgrades. If you have any suggestions on how to debug it, I'd be appreciative.

ryantm:
Closing this temporarily to stop the spam.

PR #561, opened 2021-08-12
fix scoped associations

ryantm:
Continuation of #560.

There is an SQL generation problem where a scoped association with
composite primary keys gets generated with SQL like:

```
FROM `table1` INNER JOIN `table2` ON  AND
```

ON and AND are not supposed to be next to one another.

This adds a test to uncover the issue, and then fixes it with a patch.

The problem seems to be that `append_constraints` assumes that `right.expr` is non-empty, but in the case of CPK it is sometimes empty, presumably because id constraints have been removed.

Tests passing here: https://github.com/ryantm/composite_primary_keys/pull/2/checks

### DIFF
--- a/lib/composite_primary_keys/associations/join_dependency.rb
+++ b/lib/composite_primary_keys/associations/join_dependency.rb
@@ -1,6 +1,25 @@
 module ActiveRecord
   module Associations
     class JoinDependency
+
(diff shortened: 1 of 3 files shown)


cfis:
Thanks!

PR #564, opened 2021-08-31
Set composite key autoincrement fields on create

merhard:
I have a Rails v5.1 application using the v10 version of this gem. I have run into an issue where composite keys that autoincrement do not get set on the model on create. I discovered that this feature was fixed/implemented in v11 (for Rails v5.2), but I am currently unable to upgrade the Rails version of the application for other reasons.

This PR is intended to backport functionality introduced in v11 (ActiveRecord v5.2) to v10 (ActiveRecord v5.1).

This backport for ActiveRecord v5.1 is based on the following commits in the `ar_5.2.x` branch:
- https://github.com/composite-primary-keys/composite_primary_keys/commit/1c17d72cdea4201cff2790889f266c72005af3b1
- https://github.com/composite-primary-keys/composite_primary_keys/commit/8fe118692b29cf585a3f5ab0893cc060f7d16135
- https://github.com/composite-primary-keys/composite_primary_keys/commit/902a1ae9676c4efbe251ce286469114d92bf759b
- https://github.com/composite-primary-keys/composite_primary_keys/commit/1649099fe743992773e60c5957be8db038c27275


cfis:
Thanks for taking the time to do this - and especially adding tests. The one thing that gives me pause is adding test tables in a back branch that won't be carried forward. I can live with that, but would it be better to use the tests added in 5.2?

merhard:
> would it be better to use the tests added in 5.2?

All of the tests (and fixture changes) added in this PR were backported from the `ar_5.2.x` branch commits I listed in the PR description.

cfis:
Ah, great.

PR #570, opened 2022-02-09
Follow the method name change in ActiveRecord 7.0.2

y0t4:
Fixed undefined_method error since `record_changed?` has been renamed to `_record_changed?` in ActiveRecord 7.0.2

follow this change: https://github.com/rails/rails/pull/44191


cfis:
Thanks!

PR #574, opened 2022-05-15
Accept strings in has_many ids assignment

mshibuya:
In the current implementation, we can assign has_many association using composite key ids like

```ruby
room.room_assignment_ids = [[3, 1, 2], [4, 1, 2]]
```

but it doesn't accept the string representation of composite key ids.

```ruby
room.room_assignment_ids = ['3,1,2', '4,1,2']
# => NoMethodError: undefined method `zip' for "3,1,2":String
```

This PR is an attempt to add that capability, with adding some tests.

### DIFF
--- a/lib/composite_primary_keys/associations/collection_association.rb
+++ b/lib/composite_primary_keys/associations/collection_association.rb
@@ -6,7 +6,7 @@ module CompositePrimaryKeys
 
       # CPK-
       if primary_key.is_a?(Array)
-        ids = CompositePrimaryKeys.normalize(ids, primary_key.size)
+        ids = ids.map { |id| CompositePrimaryKeys::CompositeKeys.parse(id) }
         primary_key.each_with_index do |key, i|
           pk_type = klass.type_for_attribute(key)
           ids.each { |id| id[i] = pk_type.cast(id[i]) }
@@ -15,7 +15,7 @@ module CompositePrimaryKeys
         predicate = CompositePrimaryKeys::Predicates.cpk_in_predicate(klass.arel_table, reflection.association_primary_key, ids)
         records = klass.where(predicate).index_by do |r|
           reflection.association_primary_key.map{ |k| r.send(k) }
-        end.values_at(*ids)
+        end.values_at(*ids).compact
       else
         pk_type = klass.type_for_attribute(primary_key)
         ids.map! { |i| pk_type.cast(i) }
--- a/lib/composite_primary_keys/associations/collection_association.rb
+++ b/lib/composite_primary_keys/associations/collection_association.rb
@@ -2,17 +2,24 @@ module CompositePrimaryKeys
   module CollectionAssociation
     def ids_writer(ids)
       primary_key = reflection.association_primary_key
-      pk_type = klass.type_for_attribute(primary_key)
       ids = Array(ids).reject(&:blank?)
-      ids.map! { |i| pk_type.cast(i) }
 
       # CPK-
       if primary_key.is_a?(Array)
+        ids = CompositePrimaryKeys.normalize(ids, primary_key.size)
+        primary_key.each_with_index do |primary_key, i|
+          pk_type = klass.type_for_attribute(primary_key)
+          ids.each { |id| id[i] = pk_type.cast(id[i]) }
+        end
+
         predicate = CompositePrimaryKeys::Predicates.cpk_in_predicate(klass.arel_table, reflection.association_primary_key, ids)
         records = klass.where(predicate).index_by do |r|
           reflection.association_primary_key.map{ |k| r.send(k) }
         end.values_at(*ids)
       else
+        pk_type = klass.type_for_attribute(primary_key)
+        ids.map! { |i| pk_type.cast(i) }
+
         records = klass.where(primary_key => ids).index_by do |r|
           r.public_send(primary_key)
         end.values_at(*ids).compact
--- a/lib/composite_primary_keys/associations/collection_association.rb
+++ b/lib/composite_primary_keys/associations/collection_association.rb
@@ -7,8 +7,8 @@ module CompositePrimaryKeys
       # CPK-
       if primary_key.is_a?(Array)
         ids = CompositePrimaryKeys.normalize(ids, primary_key.size)
-        primary_key.each_with_index do |primary_key, i|
-          pk_type = klass.type_for_attribute(primary_key)
+        primary_key.each_with_index do |key, i|
+          pk_type = klass.type_for_attribute(key)
           ids.each { |id| id[i] = pk_type.cast(id[i]) }
         end
 
--- a/test/test_associations.rb
+++ b/test/test_associations.rb
@@ -374,6 +374,14 @@ class TestAssociations < ActiveSupport::TestCase
     assert_equal room.room_assignments.map(&:student_id), [3, 4]
   end
 
+  def test_assignment_by_ids_as_arrays_that_contains_a_comma
+    room = Room.create dorm: dorms(:branner), room_id: 4
+    e = assert_raises ActiveRecord::RecordNotFound do
+      room.room_assignment_ids = [['5,,', '5,,', '5,,']]
+    end
+    assert_match /'student_id,dorm_id,room_id'=\[\[5, 5, 5\]\]/, e.message
+  end
+
   def test_assignment_by_ids_as_strings
     room = Room.create dorm: dorms(:branner), room_id: 4
     room.room_assignment_ids = ['3,1,2', '4,1,2']
--- a/test/test_associations.rb
+++ b/test/test_associations.rb
@@ -367,7 +367,21 @@ class TestAssociations < ActiveSupport::TestCase
     assert_equal(false, associations.send('foreign_key_present?'))
   end
 
-  def test_ids_equals_for_non_CPK_case
+  def test_assignment_by_ids_as_arrays
+    room = Room.create dorm: dorms(:branner), room_id: 4
+    room.room_assignment_ids = [[3, 1, 2], [4, 1, 2]]
+    room.save!
+    assert_equal room.room_assignments.map(&:student_id), [3, 4]
+  end
+
+  def test_assignment_by_ids_as_strings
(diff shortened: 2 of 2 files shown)


test/test_associations.rb
```diff
@@ -367,7 +367,21 @@ def test_foreign_key_present_with_null_association_ids
...
+  end
+
+  def test_assignment_by_ids_as_strings
+    room = Room.create dorm: dorms(:branner), room_id: 4
+    room.room_assignment_ids = ['3,1,2', '4,1,2']
+    room.save!
+    assert_equal room.room_assignments.map(&:student_id), [3, 4]
+  end
+
+  def test_assignment_by_ids_for_non_CPK_case
```
mshibuya, line 384: I've slightly changed the name for this test case since I felt the word `equals` might be somewhat confusing with the assertion's 'equal', but I don't have strong opinion on this.

cfis:
Thanks for the PR!

What would be the use case for accepting a string of ids? What if the ids themselves were strings and had embedded commas?

codeodor:
This could get hairy with the way we already do params[:id] being a comma delimited string. That would I suppose already fail if it included an ID that with a string that also had a comma in it.   I too would like to know more about the use case.

I don't *like* sending the params[:id] as we do and supporting splitting it in `find`, but it's so ingrained over the years that I would hate to remove it. I am happy to continue supporting that because I value backward compatibility, but I definitely lean against extending support for things like it.

cfis:
Yes, I tend to agree this could lead to unintended consequences. Make me think of removing this usage for CPK search.

Membership.find(1,1)

Because it made code to complex to deal with.

mshibuya:
First of all, I want to emphasize that this doesn't break any existing implementations. Because the current ways on how has_many id writer is used is:

|                                    | Non-array ids (`[1,2,...]`) | Array ids (`[[1,2],[1,3],...]`) |
|---------------------|--------------------------|--------------------------------|
| Non-CPK has_many | Works                                | Error (*1) | 
| CPK has_many          | Error (*2)                           | Works | 

*1 should be fine as it is, because assigning Array to single primary key is meaningless.
And what I'm talking about is the *2 case, currently if I try to do so it fails with:

```
    NoMethodError:
       undefined method `zip' for "1":String
     # /Users/mshibuya/.anyenv/envs/rbenv/versions/3.0.3/lib/ruby/gems/3.0.0/gems/composite_primary_keys-14.0.4/lib/composite_primary_keys/composite_predicates.rb:36:in `cpk_id_predicate'
     # /Users/mshibuya/.anyenv/envs/rbenv/versions/3.0.3/lib/ruby/gems/3.0.0/gems/composite_primary_keys-14.0.4/lib/composite_primary_keys/composite_predicates.rb:55:in `block in cpk_in_predicate'
     # /Users/mshibuya/.anyenv/envs/rbenv/versions/3.0.3/lib/ruby/gems/3.0.0/gems/composite_primary_keys-14.0.4/lib/composite_primary_keys/composite_predicates.rb:54:in `map'
     # /Users/mshibuya/.anyenv/envs/rbenv/versions/3.0.3/lib/ruby/gems/3.0.0/gems/composite_primary_keys-14.0.4/lib/composite_primary_keys/composite_predicates.rb:54:in `cpk_in_predicate'
     # /Users/mshibuya/.anyenv/envs/rbenv/versions/3.0.3/lib/ruby/gems/3.0.0/gems/composite_primary_keys-14.0.4/lib/composite_primary_keys/associations/collection_association.rb:11:in `ids_writer'
```

So it's safe to use this case for the enhancement, because no one depends on this behavior.

And the use case for this feature is that once has_many id writes can accept the String representation, you can pass composite ids into a form's select values:

```erb
<%= form_for @room do |f| %>
  <% # r.id.to_s => CompositeKeys string representation, like  '1,2' %>
  <%= f.select :room_assignment_ids, RoomAssignment.all.collect { |r| [ "RoomAssignment #{r.id}", r.id.to_s ] }, multiple: true %>
  <%= f.submit %>
<% end %>
```

to update the room assignments from the room side.

Above one is just a concept. My real use case is a bit complicated. This is used in the effort for making RailsAdmin compatible with CompositePrimaryKeys, which you can see in the work-in-progress commit below. 
https://github.com/railsadminteam/rails_admin/commit/a4dc93f23a6bf6a62016972154970b94c5a64478

lib/composite_primary_keys/associations/collection_association.rb
```diff
@@ -2,17 +2,24 @@ module CompositePrimaryKeys
...
     def ids_writer(ids)
       primary_key = reflection.association_primary_key
-      pk_type = klass.type_for_attribute(primary_key)
       ids = Array(ids).reject(&:blank?)
-      ids.map! { |i| pk_type.cast(i) }
 
       # CPK-
       if primary_key.is_a?(Array)
+        ids = CompositePrimaryKeys.normalize(ids, primary_key.size)
+        primary_key.each_with_index do |primary_key, i|
```
cfis, line 10: Reusing the variable primary_key in the block is confusing - it should be named something else.

lib/composite_primary_keys/associations/collection_association.rb
```diff
@@ -2,17 +2,24 @@ module CompositePrimaryKeys
...
+          pk_type = klass.type_for_attribute(primary_key)
+          ids.each { |id| id[i] = pk_type.cast(id[i]) }
```
cfis, line 12: Is this where the string is split by comma? If not, where is that code?

cfis:
Thanks @mshibuya for the additional info and working on Rails Admin.

What would be involved in approaching this from the other direction? Meaning instead of parsing '3,1,2' from form url encoded data, what if ActiveView wrote it out in the html as '[[3,1,2]]'?

lib/composite_primary_keys/associations/collection_association.rb
mshibuya: It's done [here](https://github.com/composite-primary-keys/composite_primary_keys/pull/574/files#diff-031aba8efafc35d1c9e42feddeef3a74c2acfbd8e66a8b4acb0e92d1333b527aR9):

```ruby
ids = CompositePrimaryKeys.normalize(ids, primary_key.size)
```

It uses [CompositePrimaryKeys#normalize](https://github.com/composite-primary-keys/composite_primary_keys/blob/master/lib/composite_primary_keys/composite_arrays.rb#L22) for parsing, so it supports unescaping commas originally contained in the ids.

But I realized that it's better to use [CompositePrimaryKeys::CompositeKeys#parse](https://github.com/composite-primary-keys/composite_primary_keys/blob/master/lib/composite_primary_keys/composite_arrays.rb#L41) to reduce ambiguity. #normalize's recursive behavior may lead to an unintended result, if a comma is contained in the id.

mshibuya:
> what if ActiveView wrote it out in the html as '[[3,1,2]]'?

Sorry, I didn't get this part.

With my current implementation, the rendered select tag looks like: 

```html
<select multiple="multiple" name="room[room_assignment_ids][]" id="room_room_assignment_ids">
  <option value="3,1,2">RoomAssignment #3,1,2</option>
</select>
```

and upon the form submission the Rails app will receive params like:

```
Parameters: {"room"=>{"room_assignment_ids"=>["3,1,2"]}}
```

With your proposed solution, what do they look like respectively?

cfis:
I'm thinking like this:

```
 <option value="[3,1,2]">RoomAssignment #3,1,2</option>
```

lib/composite_primary_keys/associations/collection_association.rb
cfis: Wouldn't normalize have the same problem if the keys are strings that contain commas? In this example, assume ";" is the separator"

"A key with an , comma; A second key, 5332"

Its not very likely to happen, but I'm sure someone has this. Of course, its probably wouldn't work with CPK anyway since I don't believe its ever been tested.

mshibuya: Sorry I don't understand what you're worried about.
Even if ids are String and it contain commas, the comma-separated String is generated by [CompositePrimaryKeys::CompositeKeys#to_s](https://github.com/composite-primary-keys/composite_primary_keys/blob/master/lib/composite_primary_keys/composite_arrays.rb#L54) anyway. It escapes potentially conflicting characters, so there'll be no issue.

mshibuya:
What's the benefit of that?
My understanding is that this comma-separated format is the standard String representation of the [CompositePrimaryKeys::CompositeKeys](https://github.com/composite-primary-keys/composite_primary_keys/blob/master/lib/composite_primary_keys/composite_arrays.rb#L39). Does it mean you want  to create another String representation?

lib/composite_primary_keys/associations/collection_association.rb
cfis: Good point. I forgot about https://github.com/composite-primary-keys/composite_primary_keys/pull/506.

cfis:
To flag it as a CPK key so there is no confusion. Now that you mention it, maybe it should have been done that way from the very beginning.

cfis:
So why would this change only impact collection associations? Why not other types of associations?

mshibuya:
Because what I'm working on here is [ActiveRecord::Associations::CollectionAssociation#ids_writer](https://github.com/rails/rails/blob/75a9e1be75769ae633a938d81d51e06852a69ea3/activerecord/lib/active_record/associations/collection_association.rb#L60). So only descendants of `CollectionAssociation` (= `HasManyAssociation` and `HasManyThroughAssociation`) are affected. 

Also the composite key parse logic is guarded by this conditional:

```ruby
      if primary_key.is_a?(Array)
```

https://github.com/composite-primary-keys/composite_primary_keys/pull/574/files#diff-031aba8efafc35d1c9e42feddeef3a74c2acfbd8e66a8b4acb0e92d1333b527aR8

So the CPK parse will never be executed unless the association target class has composite primary keys.

cfis:
Ok, let's give this a try then. Thanks for your answers.

mshibuya:
Thank you for merging, along with giving the attentive review 👍

Issue #558, opened 2021-06-03
ArgumentError: Polymorphic association does not support to compute class


mediafinger:
Thank you for maintaining this library 😃   
We only encountered it now, after having a composite index in one of our MySQL tables and threw it in the mix. Surprisingly this leads to multiple errors in unrelated classes:

The call to `source_reflection.klass` will raise the ArgumentError:  
`Polymorphic association does not support to compute class`   
when `source_reflection.klass` is polymorphic.

```ruby
module ActiveRecord
  module Associations
    module ThroughAssociation
      alias :original_construct_join_attributes :construct_join_attributes

      def construct_join_attributes(*records)
        # CPK
        if source_reflection.klass.composite?
          ensure_mutable

          ids = records.map do |record|
            source_reflection.association_primary_key(reflection.klass).map do |key|
              record.send(key)
            end
          end

          cpk_in_predicate(through_association.scope.klass.arel_table, source_reflection.foreign_key, ids)
        else
          original_construct_join_attributes(*records)
        end
      end
    end
  end
end
```

This change was discussed here: https://github.com/rails/rails/pull/31877  
and implemented here: https://github.com/rails/rails/commit/fb86ecd604a362355827bbf05776b847b0ded9a5

For our code base a monkey-patch to change the condition to:  
`if !source_reflection.polymorphic? && source_reflection.klass.composite?`  
would suffice, but I doubt this would be a good solution for this gem.

cfis:
I'm ok with that change. Would you be able to submit a patch with a test case?

cfis:
After looking at this again, I think your patch is good enough for now and I applied it to the Rails 6.1 and Rails 7 branches. Thanks!

PR #575, opened 2022-06-27
fix reflection with polymorphic association

hassan-ahmed:
Hi, thanks for the wonderful gem.
I stumbled upon this error, that's already solved for rails 6.x & 7.x, but since we're still on rails 5.x, I thought of making the fix.

Fixes the issue described in https://github.com/composite-primary-keys/composite_primary_keys/issues/558, for rails 5.2x


cfis:
Sorry for the delay on this one.

PR #586, opened 2022-12-05
ar_6.1.x: Fix previously_new_record? not being set to true after create

knu:
It was introduced in Rails 6.1.0, so composite-primary-keys should support it in the 13.x series.
https://api.rubyonrails.org/v6.1.0/classes/ActiveRecord/Persistence.html#method-i-previously_new_record-3F


cfis:
Thanks!

PR #589, opened 2023-01-14
Adds Ruby 3.2 to CI.  Updates checkout action versions.

petergoldstein:
Also quotes "3.0" to avoid truncation.  An unquoted "3.0" is truncated to "3", which causes the latest Ruby 3 (at this point 3.2.0) to be loaded, which is not what's intended.

This runs green on my fork.


cfis:
Thanks!

PR #591, opened 2023-01-25
Improve query generation for cpk_in_predicate

akiellor:
Query generation for cpk_in_predicate now constructs queries in the form:

```sql
SELECT *
FROM table_name
WHERE low_cardinality_key_part = 1 AND high_cardinality_part IN (1, 2)
```

It used to generate queries in the form:

```sql
SELECT *
FROM table_name
WHERE
  (low_cardinality_key_part = 1 AND high_cardinality_part = 1)
  OR (low_cardinality_key_part = 1 AND high_cardinality_part = 2)
```

This change improves the queries by reducing the overall length of the query, especially when loading many keys. But more importantly the new query will often result in Postgres performing an `Index Scan` instead of a `Bitmap Heap Scan` (assuming the right indices have been added).

### DIFF
--- a/lib/composite_primary_keys/composite_predicates.rb
+++ b/lib/composite_primary_keys/composite_predicates.rb
@@ -51,9 +51,43 @@ module CompositePrimaryKeys
     end
 
     def cpk_in_predicate(table, primary_keys, ids)
-      and_predicates = ids.map do |id|
-        cpk_id_predicate(table, primary_keys, id)
+      if primary_keys.length == 2
+        low_cardinality_key_part, high_cardinality_key_part = if ids.map(&:second).uniq.size > ids.map(&:first).uniq.size
+                                                                %i[first second]
+                                                              else
+                                                                %i[second first]
+                                                              end
+
+        groups = ids.group_by(&low_cardinality_key_part).transform_values do |values|
+          values.map(&high_cardinality_key_part)
+        end
+
+        and_predicates = groups.map do |low_cardinality_value, high_cardinality_values|
+          in_clause = table[primary_keys.send(high_cardinality_key_part)].in(high_cardinality_values.compact)
+          inclusion_clauses = if high_cardinality_values.include?(nil)
+                                Arel::Nodes::Grouping.new(
+                                  Arel::Nodes::Or.new(
+                                    in_clause,
+                                    table[primary_keys.send(high_cardinality_key_part)].eq(nil)
+                                  )
+                                )
+                              else
+                                in_clause
+                              end
+
+          Arel::Nodes::And.new(
+            [
+              table[primary_keys.send(low_cardinality_key_part)].eq(low_cardinality_value),
+              inclusion_clauses
+            ]
+          )
+        end
+      else
+        and_predicates = ids.map do |id|
+          cpk_id_predicate(table, primary_keys, id)
+        end
       end
+
       cpk_or_predicate(and_predicates)
     end
   end
--- a/lib/composite_primary_keys/composite_predicates.rb
+++ b/lib/composite_primary_keys/composite_predicates.rb
@@ -70,9 +70,9 @@ module CompositePrimaryKeys
       keys_by_first_column_name = Hash.new { |hash, key| hash[key] = [] }
       keys_by_second_column_name = Hash.new { |hash, key| hash[key] = [] }
 
-      ids.map.each do |first_column_name, second_column_name|
-        keys_by_first_column_name[first_column_name] << second_column_name
-        keys_by_second_column_name[second_column_name] << first_column_name
+      ids.map.each do |first_key_part, second_key_part|
+        keys_by_first_column_name[first_key_part] << second_key_part
+        keys_by_second_column_name[second_key_part] << first_key_part
       end
 
       low_cardinality_column_name, high_cardinality_column_name, groups = \
--- a/lib/composite_primary_keys/composite_predicates.rb
+++ b/lib/composite_primary_keys/composite_predicates.rb
@@ -83,8 +83,9 @@ module CompositePrimaryKeys
         end
 
       and_predicates = groups.map do |low_cardinality_value, high_cardinality_values|
-        in_clause = table[high_cardinality_column_name].in(high_cardinality_values.compact)
-        inclusion_clauses = if high_cardinality_values.include?(nil)
+        non_nil_high_cardinality_values = high_cardinality_values.compact
+        in_clause = table[high_cardinality_column_name].in(non_nil_high_cardinality_values)
+        inclusion_clauses = if non_nil_high_cardinality_values.size != high_cardinality_values.size
                               Arel::Nodes::Grouping.new(
                                 Arel::Nodes::Or.new(
                                   in_clause,
--- a/lib/composite_primary_keys/composite_predicates.rb
+++ b/lib/composite_primary_keys/composite_predicates.rb
@@ -67,28 +67,28 @@ module CompositePrimaryKeys
     end
 
     def cpk_in_predicate_with_grouped_keys(table, primary_keys, ids)
-      keys_by_first_key_part = Hash.new { |h, k| h[k] = [] }
-      keys_by_second_key_part = Hash.new { |h, k| h[k] = [] }
+      keys_by_first_column_name = Hash.new { |hash, key| hash[key] = [] }
+      keys_by_second_column_name = Hash.new { |hash, key| hash[key] = [] }
 
-      ids.map.each do |first_key_part, second_key_part|
-        keys_by_first_key_part[first_key_part] << second_key_part
-        keys_by_second_key_part[second_key_part] << first_key_part
+      ids.map.each do |first_column_name, second_column_name|
+        keys_by_first_column_name[first_column_name] << second_column_name
+        keys_by_second_column_name[second_column_name] << first_column_name
       end
 
-      low_cardinality_key_part, high_cardinality_key_part, groups = \
-        if keys_by_first_key_part.keys.size <= keys_by_second_key_part.keys.size
-          primary_keys + [keys_by_first_key_part]
+      low_cardinality_column_name, high_cardinality_column_name, groups = \
+        if keys_by_first_column_name.size <= keys_by_second_column_name.size
+          primary_keys + [keys_by_first_column_name]
         else
-          primary_keys.reverse + [keys_by_second_key_part]
+          primary_keys.reverse + [keys_by_second_column_name]
         end
 
       and_predicates = groups.map do |low_cardinality_value, high_cardinality_values|
-        in_clause = table[high_cardinality_key_part].in(high_cardinality_values.compact)
+        in_clause = table[high_cardinality_column_name].in(high_cardinality_values.compact)
         inclusion_clauses = if high_cardinality_values.include?(nil)
                               Arel::Nodes::Grouping.new(
                                 Arel::Nodes::Or.new(
                                   in_clause,
-                                  table[high_cardinality_key_part].eq(nil)
+                                  table[high_cardinality_column_name].eq(nil)
                                 )
                               )
                             else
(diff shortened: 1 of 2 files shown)


codeodor:
Does it handle it in the case of having `low_cardinality_part` with a count more than 1? 

i.e., 

`low_cardinality_part = 1 and high_cardinality_part = 1 or low_cardinality_part = 1 and high_cardinality_part = 2 or low_cardinality_part = 2 and high_cardinality_part = 3`

is not the same as 

`low_cardinality_part in (1, 2) and  high_cardinality_part in (1, 2, 3)` if we have a table like:

```
low   high
1      1
1      2 
2      3
2      1

```

It would pull (2,1) in the latter case but not the former.

akiellor:
@codeodor it does, but there is no spec. I'll add one.

akiellor:
@codeodor Actually there is a [spec](https://github.com/composite-primary-keys/composite_primary_keys/pull/591/files#diff-e2fe044a18e7088d4ef8ca4cf3eefbd85c8a0a357fd7bdb08a0a20a691518f6aR103).

```
  def test_in_with_multiple_primary_key_parts
    dep = Department.arel_table

    primary_keys = [[1, 1], [1, 2], [2, 3], [2, 4]]

    connection = ActiveRecord::Base.connection
    quoted_id_column = "#{connection.quote_table_name('departments')}.#{connection.quote_column_name('id')}"
    quoted_location_id_column = "#{connection.quote_table_name('departments')}.#{connection.quote_column_name('location_id')}"
    expected = "(#{quoted_id_column} = 1 AND #{quoted_location_id_column} IN (1, 2) OR #{quoted_id_column} = 2 AND #{quoted_location_id_column} IN (3, 4))"

    pred = cpk_in_predicate(dep, [:id, :location_id], primary_keys)
    assert_equal(with_quoted_identifiers(expected), pred.to_sql)
  end
  ```
  
  Note how the query ends up being structured in a way where the high cardinality parts are grouped by the low cardinality parts and `OR`ed together.

File: lib/composite_primary_keys/composite_predicates.rb [hunk 1]
```diff
@@ -51,9 +51,43 @@ def cpk_join_predicate(table1, key1, table2, key2)
     end
 
     def cpk_in_predicate(table, primary_keys, ids)
-      and_predicates = ids.map do |id|
-        cpk_id_predicate(table, primary_keys, id)
+      if primary_keys.length == 2
+        low_cardinality_key_part, high_cardinality_key_part = if ids.map(&:second).uniq.size > ids.map(&:first).uniq.size
```
cfis, line 55: This loops over the ids twice which may impact performance with a lot of ids. Probably better to have a single loop?

Also I wonder if having a hash table keyed on id with the values the count of each item would perform better? Avoids the uniq calls. Would guess that would be better if there are a lot of duplicates and worse if they aren't.\

lib/composite_primary_keys/composite_predicates.rb
```diff
@@ -51,9 +51,43 @@ def cpk_join_predicate(table1, key1, table2, key2)
...
+                                                                %i[first second]
+                                                              else
+                                                                %i[second first]
+                                                              end
+
+        groups = ids.group_by(&low_cardinality_key_part).transform_values do |values|
```
cfis, line 61: Since we are actually already creating hash tables keyed on id anyway (but with the 2nd column values instead), why not just do this at the start?

```
hash1 = Hash.new
hash2 = Hash.n32
ids.map.each do |value1, value2|
  hash1[value1] << value2
  hash2[value2] << value1
end

Then compare the number of keys in hash1 versus hash2 and dump the one with more keys. 

```

lib/composite_primary_keys/composite_predicates.rb
```diff
@@ -51,9 +51,43 @@ def cpk_join_predicate(table1, key1, table2, key2)
...
+
+          Arel::Nodes::And.new(
+            [
+              table[primary_keys.send(low_cardinality_key_part)].eq(low_cardinality_value),
+              inclusion_clauses
+            ]
+          )
+        end
+      else
+        and_predicates = ids.map do |id|
```
cfis, line 86: I'd probably split this method into 3 methods:

  cpk_in_predicate
  cpk_in_predicate_pairs
  cpk_in_predicate_tuples (not a very good name, have a better idea?)

akiellor:
Thanks for the feedback y'all.

@cfis,

I've refactored the code a little, let me know what you think.

It was definitely possible to do all the work with a single pass of the list. I honestly hadn't given it much thought as the actual query in Postgres was dominating the end to end execution, but this is better.

I went with some different method names, different to your suggestions, but hopefully reflects the intent.

lib/composite_primary_keys/composite_predicates.rb
```diff
@@ -52,40 +52,55 @@ def cpk_join_predicate(table1, key1, table2, key2)
...
+    def cpk_in_predicate_with_non_grouped_keys(table, primary_keys, ids)
+      and_predicates = ids.map do |id|
+        cpk_id_predicate(table, primary_keys, id)
+      end
+
+      cpk_or_predicate(and_predicates)
+    end
+
+    def cpk_in_predicate_with_grouped_keys(table, primary_keys, ids)
+      keys_by_first_key_part = Hash.new { |h, k| h[k] = [] }
```
cfis, line 70: Nitpick. Can this be Hash.new { |hash, key| hash[key] = [] }?

lib/composite_primary_keys/composite_predicates.rb
```diff
@@ -52,40 +52,55 @@ def cpk_join_predicate(table1, key1, table2, key2)
...
+      keys_by_second_key_part = Hash.new { |h, k| h[k] = [] }
+
+      ids.map.each do |first_key_part, second_key_part|
+        keys_by_first_key_part[first_key_part] << second_key_part
+        keys_by_second_key_part[second_key_part] << first_key_part
+      end
+
+      low_cardinality_key_part, high_cardinality_key_part, groups = \
+        if keys_by_first_key_part.keys.size <= keys_by_second_key_part.keys.size
```
cfis, line 79: Simpler - `keys_by_first_key_part.size`

lib/composite_primary_keys/composite_predicates.rb
```diff
@@ -52,40 +52,55 @@ def cpk_join_predicate(table1, key1, table2, key2)
...
-                                                              else
-                                                                %i[second first]
-                                                              end
-
-        groups = ids.group_by(&low_cardinality_key_part).transform_values do |values|
-          values.map(&high_cardinality_key_part)
+        cpk_in_predicate_with_grouped_keys(table, primary_keys, ids)
+      else
+        cpk_in_predicate_with_non_grouped_keys(table, primary_keys, ids)
+      end
```
cfis, line 58: I like these method names better!

lib/composite_primary_keys/composite_predicates.rb
```diff
@@ -52,40 +52,55 @@ def cpk_join_predicate(table1, key1, table2, key2)
...
+      keys_by_second_key_part = Hash.new { |h, k| h[k] = [] }
+
+      ids.map.each do |first_key_part, second_key_part|
+        keys_by_first_key_part[first_key_part] << second_key_part
+        keys_by_second_key_part[second_key_part] << first_key_part
+      end
+
+      low_cardinality_key_part, high_cardinality_key_part, groups = \
+        if keys_by_first_key_part.keys.size <= keys_by_second_key_part.keys.size
+          primary_keys + [keys_by_first_key_part]
```
cfis, line 80: I'm confused by this. Why is primary keys being added to the hash table?

cfis:
I like this better - thanks for the changes although I'm sure you a right that the query building and execution is where all the time is spent.

Added a second round of comments.

lib/composite_primary_keys/composite_predicates.rb
akiellor: Done.

lib/composite_primary_keys/composite_predicates.rb
akiellor: Done.

lib/composite_primary_keys/composite_predicates.rb
akiellor: This code is making an array where the elements are `[primary_keys.first, primary_keys.second, groups]` (not modifying the hash). When keys_by_second_column_name has fewer keys the returned array is `[primary_keys.second, primary_keys.first, groups]`. This is important to make sure that `low_cardinality_column_name` and `high_cardinality_column_name` are bound to the right column_names.

I'm thinking that maybe explicitly defining the arrays rather than calculating them might be more clear. I'll push that up and you can see what you think.

akiellor:
@cfis, I think this is ready for another round. Thanks for your feedback.

lib/composite_primary_keys/composite_predicates.rb
```diff
@@ -51,9 +51,58 @@ def cpk_join_predicate(table1, key1, table2, key2)
...
+
+      low_cardinality_column_name, high_cardinality_column_name, groups = \
+        if keys_by_first_column_name.size <= keys_by_second_column_name.size
+          [primary_keys.first, primary_keys.second, keys_by_first_column_name]
+        else
+          [primary_keys.second, primary_keys.first, keys_by_second_column_name]
+        end
+
+      and_predicates = groups.map do |low_cardinality_value, high_cardinality_values|
+        in_clause = table[high_cardinality_column_name].in(high_cardinality_values.compact)
```
cfis, line 86: postgresql, mysql, oracle, etc do not allow null values in primary keys (in any field). So the compact call should be unnecessary as well as the include?(nil) check.

akiellor: I'm happy to remove it. But I think cpk can be used in situations where the columns defined in the ActiveRecord primary key are not the primary key in the database.

We do this in some situations (e.g. views), but we always require both parts to be present. I'm happy to remove this, if that is desirable.

codeodor:
@akiellor nice! Thanks for pointing that out to me. I missed it on my first pass.

lib/composite_primary_keys/composite_predicates.rb
cfis: What happens in the current implementation with nil values? If they aren't supported now I would not add support for them now.

akiellor: The existing implementation of cpk_in_predicate (or in this PR cpk_in_predicate_with_non_grouped_keys) would create a queries like:

```sql
# pk [1, 1]
SELECT * FROM table WHERE pk1 = 1 AND pk2 = 1

# pk [1, nil]
SELECT * FROM table WHERE pk1 = 1 AND pk2 IS NULL

# pk [nil, 1]
SELECT * FROM table WHERE pk1 IS NULL AND pk2 = 1

# pk [nil, nil]
SELECT * FROM table WHERE pk1 IS NULL AND pk2 IS NULL
```

Arel transparently converts something like `table[column_name].eq(value)` to an `IS NULL` if the value is `nil`.

cfis: Ah - then let's keep the null handling then.

cfis: Ah - then let's keep the null handling then.

cfis: However, the code checks the array twice for nil, once with `compact` and once with `'include?`. Let's just check once.

```
compacted = high_cardinality_values.compact
...
if high_cardinality_values.size > compacted.size
...
```

akiellor: Done, feel free to take a look.

cfis:
Great - thanks for making this happen!

akiellor:
@cfis, is this change likely to be backported to all the different ActiveRecord version releases?

cfis:
Happy to merge a MR if you want to make one. 

Also, I'll cut a new master branch release in the next few days.

akiellor:
I'm not sure of the exact process, I did just open a new [PR](https://github.com/composite-primary-keys/composite_primary_keys/pull/592) into the ar_6.1.x branch with a the commits squashed from the previous PR.

PR #592, opened 2023-02-04
AR 6.1 - Improve query generation for cpk_in_predicate

akiellor:
Query generation for cpk_in_predicate now constructs queries in the form:

```
SELECT *
FROM table_name
WHERE low_cardinality_key_part = 1 AND high_cardinality_part IN (1, 2)
```

It used to generate queries in the form:

```
SELECT *
FROM table_name
WHERE
  (low_cardinality_key_part = 1 AND high_cardinality_part = 1)
  OR (low_cardinality_key_part = 1 AND high_cardinality_part = 2)
```

This change improves the queries by reducing the overall length of the query, especially when loading many keys. But more importantly the new query will often result in Postgres performing an `Index Scan` instead of a `Bitmap Heap Scan` (assuming the right indices have been added).

### DIFF
--- a/lib/composite_primary_keys/composite_predicates.rb
+++ b/lib/composite_primary_keys/composite_predicates.rb
@@ -51,9 +51,59 @@ module CompositePrimaryKeys
     end
 
     def cpk_in_predicate(table, primary_keys, ids)
+      if primary_keys.length == 2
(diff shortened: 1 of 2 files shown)


cfis:
Thanks!

PR #597, opened 2023-05-06
Fix HasOneThrough association

kyori19:
Installing this gem breaks ActiveRecord `HasOneThrough` association.
It is due to overriding `construct_join_attributes` in `ThroughAssociation`.

Overriding `construct_join_attributes` helps `HasManyThroughAssociation` to find by multiple set of IDs on composite primary key.
This is not needed for `HasOneThroughAssociation` because it won't be associated to multiple records.

Therefore, I moved codes to override `construct_join_attributes` to `HasManyThroughAssociation` and added tests to ensure `HasOneThrough` association correctly works.


cfis:
Thanks @kyori19 for your work on this. Looks like the commit also makes lots of changes to database tables (updating DB2 and Oracle).

Could you remove those changes from this commit and create a new one for them? So this commit focuses just on FixHasOne Through associates. If this commit depends on those changes, then let's merge those first.

kyori19:
@cfis I tried to remove all changes not related to the PR title but made automatically by the editor. Furthermore, I can remove entire changes to the DB2 and Oracle schemas which is not tested at all, if you want.

By the way, do you want me to make another PR to remove trailing spaces from those SQLs?

cfis:
Thanks for the updates. Any chance we can *not* introduce new tables for testing and use existing ones? I've been trying hard to remove the number of test tables which has gotten a bit out of hand in my view?

I was thinking on MR with the db changes and one MR with the Fix HasOneThrough association fixes.

cfis:
Ok. Thanks for doing this!

PR #603, opened 2023-09-08
AR 5.2: Support polymorphic_name

marshall-lee:
`polymorphic_name` was introduced in 5.2.0 but wasn't properly ported to CPK.

### DIFF
--- a/.github/workflows/mariadb.yml
+++ b/.github/workflows/mariadb.yml
@@ -1,0 +1,45 @@
+name: mariadb
+
+on: [push, pull_request]
+
+jobs:
+  mariadb_test:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+        ruby: [2.5, 2.6, 2.7]
+    runs-on: ${{matrix.os}}
+    services:
+      mariadb:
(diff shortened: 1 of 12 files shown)


marshall-lee:
@cfis Hi and thanks for merging of #604 #605 and #606 to support `polymorphic_name` in Rails >= 6!
But one of my codebases is still on 5.2 and this fix is actually crucial there. I humbly ask you please merge the fix into upstream so I can refer to the official repo in my `Gemfile`. I know 5.2 is EOL many years ago but things are not that smooth sometimes, sometimes codebases are stuck with older versions.

cfis:
Yes sure, thought I had merged. Done now.

marshall-lee:
@cfis Hi! Is there a chance `11.x`, `12.x` and `13.x` versions with this fix be released on rubygems?

cfis:
Well I would usually only do 14.x and 13.x  - anything older is unsupported. Do you really need it on all those releases?

marshall-lee:
@cfis i'm in the process of upgrading to 6.1 so 13.x would be fine!

cfis:
Coo, done.

PR #606, opened 2023-09-08
Support polymorphic_name

marshall-lee:
`polymorphic_name` was introduced in 5.2.0 but wasn't properly ported to CPK.


cfis:
Thanks!

PR #607, opened 2023-09-12
Feature: support trilogy adapter

zmariscal:
# Purpose
There's a new DB adapter called [trilogy](https://github.blog/2022-08-25-introducing-trilogy-a-new-database-adapter-for-ruby-on-rails/) and this gem should be aware of it.

# Implementation
Make the gem aware of the trilogy adapter by adding the gem and subsequent requires.  We need to make sure the database statements `value` is actually and `ActiveRecord::Result` and not a `Trilogy::Result` when the adapter is `trilogy`. One of the reasons we need to care about this is because `Trilogy::Result` doesn't have `#empty?` implemented.


lib/composite_primary_keys/relation.rb
```diff
@@ -1,5 +1,7 @@
 module ActiveRecord
   class Relation
+    AR_CONNECTION_TRILOGY_ADAPTER = ActiveRecord::ConnectionAdapters::TrilogyAdapter
```
marshall-lee, line 3: That will fail if constant is not defined.

lib/composite_primary_keys.rb
```diff
@@ -61,6 +61,8 @@
 require 'active_record/connection_adapters/abstract/database_statements'
 require 'active_record/connection_adapters/abstract_adapter'
 require 'active_record/connection_adapters/postgresql/database_statements'
+require "active_record/connection_adapters/trilogy_adapter"
+require "active_record/connection_adapters/trilogy/database_statements"
```
cfis, line 65: These two requires should be removed. The only reason to have them is to enable CPK to override code contained in them, but I see no such overrides. In addition, these requires would make everyone using CPK have to install the trilogy adapter which we definitely don't want to do.

lib/composite_primary_keys.rb
```diff
@@ -114,4 +116,4 @@
 
 require_relative 'composite_primary_keys/arel/to_sql'
 require_relative 'composite_primary_keys/arel/sqlserver'
-require_relative 'composite_primary_keys/table_metadata'
\ No newline at end of file
+require_relative 'composite_primary_keys/table_metadata'
```
cfis, line 119: This looks like a needless change - since in the end there is no change.

lib/composite_primary_keys/connection_adapters/abstract/database_statements.rb
```diff
@@ -32,6 +31,16 @@ def insert(arel, name = nil, pk = nil, id_value = nil, sequence_name = nil, bind
...
       end
+
+      private
+
+      def active_record_result(value)
+        return value if @config[:adapter] != "trilogy"
+
+        # Create an AR Result so we can respond to #empty?
+        # since Trilogy:Result doesn't respond to it.
+        ActiveRecord::Result.new(value.fields, value.to_a)
```
cfis, line 42: Hmm, this seems like a pretty aggressive change that could have unforeseen consequences. Maybe add `empty?` to the Trilogy::Result instead? Or if trilogy just skip the check?

lib/composite_primary_keys/relation.rb
cfis: Right, this assumes Trilogy adapter is loaded which will not be the case the vast majority of times.

lib/composite_primary_keys/relation.rb
```diff
@@ -102,6 +104,8 @@ def cpk_subquery(stmt)
       # database adapter to decide how to proceed.
       if defined?(ActiveRecord::ConnectionAdapters::Mysql2Adapter) && connection.is_a?(ActiveRecord::ConnectionAdapters::Mysql2Adapter)
         cpk_mysql_subquery(stmt)
+      elsif defined?(AR_CONNECTION_TRILOGY_ADAPTER) && connection.is_a?(AR_CONNECTION_TRILOGY_ADAPTER)
+        cpk_mysql_subquery(stmt)
```
cfis, line 108: Just follow the same pattern the MySQL check uses.

test/abstract_unit.rb
```diff
@@ -1,6 +1,11 @@
 spec_name = ENV['ADAPTER'] || 'postgresql'
 require 'bundler'
 require 'minitest/autorun'
+if spec_name == "trilogy"
+  require "activerecord-trilogy-adapter"
+  require "trilogy_adapter/connection"
+  ActiveRecord::Base.extend TrilogyAdapter::Connection
+end
```
cfis, line 8: Let's remove this code. Loading the gems should be done in the Gemfile not here.

Is extending ActiveRecord::Base really needed? If so, then doesn't this has to happen when loading the trilogy gem in all cases (not just tests)?

test/connections/databases.ci.yml
```diff
@@ -20,3 +20,13 @@ sqlite:
   adapter: sqlite3
   database: <%= File.join(project_root, 'db', 'composite_primary_keys_unittest.sqlite') %>
 
+trilogy:
+  adapter: mysql2
```
cfis, line 24: The adapter isn't trilogy?

cfis:
Thanks for working on this! Had a chance to review the code, and overall looks pretty good. I did leave some comments however if you could take a look.

test/connections/databases.ci.yml
zmariscal: You're right. Not sure why I had this. I'll use trilogy 😅

cfis:
FYI I enabled all tests for this MR - right now they all fail due to requiring trilogy.

lib/composite_primary_keys/connection_adapters/abstract/database_statements.rb
zmariscal: @cfis are you suggesting adding #empty? to the trilogy gem? If so, I actually opened a PR to do that, however the maintainers didn't see it necessary as it's not meant to behave like a AR Result.

lib/composite_primary_keys.rb
zmariscal: This was my editor just adding a new line at the end of the file. Happy to remove it.

lib/composite_primary_keys/connection_adapters/abstract/database_statements.rb
zmariscal: I was thinking of adding something like this below line 8:
```ruby
return value if @config[:adapter] == "trilogy"
```

But open for suggestions on how you'd best handle this.

cfis:
Progress - now tests pass again except the Trilogy ones. Looks like a problem loading trilogy:

> 
/opt/hostedtoolcache/Ruby/3.2.2/x64/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require': Could not load the 'trilogy' Active Record adapter. Ensure that the adapter is spelled correctly in config/database.yml and that you've added the necessary adapter gem to your Gemfile. (LoadError)

zmariscal:
@cfis pushed up some commit that address the `#empty?` call for Trilogy.  I'm thinking why the Trilogy tests are failing are around how AR when initially loaded isn't aware of Trilogy and that's why we need to require them again and then load. So adding the requires and connecting via `ActiveRecord::Base.extend TrilogyAdapter::Connection` seems to have everything working.

Screenshot 2023-09-25 at 8 57 29 PM
Screenshot 2023-09-25 at 8 57 47 PM

cfis:
Still failing, with a slightly different error now:

zmariscal:
> Still failing, with a slightly different error now: 
> > database configuration specifies nonexistent trilogy adapter (ActiveRecord::AdapterNotFound)

I mentioned in my previous  comment that I need to ensure AR is aware of Trilogy. Do you mind kick off another round of the CI?

cfis:
And onto the next issue:

cfis:
And onto the next issue:

> home/runner/work/composite_primary_keys/composite_primary_keys/vendor/bundle/ruby/3.1.0/gems/trilogy-2.6.0/lib/trilogy.rb:16:in `_initialize': trilogy_auth_recv: TRILOGY_UNEXPECTED_PACKET (Trilogy::QueryError)

cfis:
Thoughts?

zmariscal:
> Thoughts?

Apologies for the delay. Let me look take a look and see.

zmariscal:
> Thoughts?
@cfis I pushed a commit as this may be an issue with Github Actions. Here's a [link](https://github.com/trilogy-libraries/activerecord-trilogy-adapter/issues/64) to the issue. 

I setup a docker file and docker-compose.yml and everything seems to be working. Happy to open a PR to commit those to the project if it helps debug this issue.

.github/workflows/trilogy.yml
```diff
@@ -43,8 +43,8 @@ jobs:
           MYSQL_ROOT_PASSWORD: root
       - name: Create user, grant privledges,and flush # workaround to resolve https://github.com/trilogy-libraries/activerecord-trilogy-adapter/issues/64
         run: |
-          mysql -e "CREATE USER 'test'@'localhost' IDENTIFIED WITH mysql_native_password BY 'test';" -u${{ env.MYSQL_USER }} -p${{ env.MYSQL_PASSWORD }}
-          mysql -e "GRANT ALL PRIVILEGES ON *.* TO 'test'@'localhost' WITH GRANT OPTION;" -u${{ env.MYSQL_USER }} -p${{ env.MYSQL_PASSWORD }}
-          mysql -e "FLUSH PRIVILEGES;" -u${{ env.MYSQL_USER }} -p${{ env.MYSQL_PASSWORD }}
+          mysql -h 127.0.0.1 -e "CREATE USER 'test'@'localhost' IDENTIFIED WITH mysql_native_password BY 'test';" -u${{ env.MYSQL_USER }} -p${{ env.MYSQL_PASSWORD }}
+          mysql -h 127.0.0.1 -e "GRANT ALL PRIVILEGES ON *.* TO 'test'@'localhost' WITH GRANT OPTION;" -u${{ env.MYSQL_USER }} -p${{ env.MYSQL_PASSWORD }}
+          mysql -h 127.0.0.1 -e "FLUSH PRIVILEGES;" -u${{ env.MYSQL_USER }} -p${{ env.MYSQL_PASSWORD }}
```
zmariscal, line 48: @cfis do you know if there's a way for me to run the GitHub actions on my own fork? I'm making educated guesses at the Github action issues around these mysql commands.

cfis: I think via the .github/workflows/trilogy.yml file. But since that is part of the PR maybe it doesn't work (might have to be merged first?)

zmariscal: @cfis I can spin up a new PR and see if merging in other components first resolve this issue.

zmariscal: @cfis any objections in me breaking this into two PR's?

cfis: Works for me. Sorry this taking so long to merge. I can try and focus on it next week.

zmariscal: First PR is here: https://github.com/composite-primary-keys/composite_primary_keys/pull/610

zmariscal: I see two more PRs after that one. The first with the trilogy guts and the second being the GitHub action. How do you feel about that?

zmariscal:
@cfis I've go ahead and removed the Github Action for now. You should be able to run the `docker compose up` command and then `docker compose exec app bash`. You may need to do some DB setup, but ultimately once everything is hooked up correctly you should be able to execute `bundle exec rake trilogy:test` and get the following [result](https://imgur.com/a/2Psg1YX).

cfis:
Sorry for the delay on this and thanks for keeping at it!

PR #610, opened 2023-10-21
Add docker and docker compose files

zmariscal:
# Purpose
To make development more convenient for others adding a Dockerfile and compose file seems reasonable. This way if/when an issue arises this will give everyone a place where they can rally to offer debugging help, etc.

# Implementation
- Add a Dockerfile 
- Add docker-compose.yml

# Demo
Screenshot 2023-10-21 at 1 55 04 PM

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,21 @@
+# Use the official Ruby 3.2 image as a base image
+ARG RUBY_VERSION=3.2
+FROM ruby:${RUBY_VERSION}-bullseye
+
+# Set the working directory
+WORKDIR /usr/src/app
+
+# Install system packages
+RUN apt-get update -qq && \
+    apt-get install -y default-mysql-client postgresql postgresql-contrib vim && \
+    apt-get clean
+
+# Copy all files
+COPY . .
+
+RUN bundler config set --local without "db2 oracle sqlserver sqlite postgresql"
+# Move sample database.yml and install gems
+RUN mv test/connections/databases.example.yml test/connections/databases.yml && \
+    bundle install
+
+CMD ["irb"]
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,21 @@
+# Use the official Ruby 3.2 image as a base image
+ARG RUBY_VERSION=3.2
```
zmariscal, line 2: I figure if someone is using a particular version of Ruby they can change change it here.

Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+FROM ruby:${RUBY_VERSION}-bullseye
+
+# Set the working directory
+WORKDIR /usr/src/app
+
+# Install system packages
+RUN apt-get update -qq && \
+    apt-get install -y default-mysql-client postgresql postgresql-contrib vim && \
```
zmariscal, line 10: I'm installing `postgresql` too. With the hope I'll have a future PR where I can add it via `docker-compose.yml`. It doesn't hurt anything currently to leave it in. I'm also installing `vim` because someone may want to edit files inside the container.

Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+    apt-get clean
+
+# Copy all files
+COPY . .
+
+RUN bundler config set --local without "db2 oracle sqlserver sqlite postgresql"
```
zmariscal, line 16: If/when we want this to work out of the box with `postgresql` this would need to be removed.

cfis:
Cool - thank!

zmariscal:
@cfis is there anything else needed for this PR? I'm waiting for this one to merge so I can create a couple of other PRs.

cfis:
Oh sorry, thought I had merged it.

PR #611, opened 2023-11-06
Add trilogy GitHub workflow

zmariscal:
# Purpose
To ensure nothing breaks the trilogy adapter we want there to be a workflow that runs tests against it. So add a GitHub workflow.

# Implementation
I attempted to add this here https://github.com/composite-primary-keys/composite_primary_keys/pull/607 but couldn't get it to work for some reason. When looking at why it works in the docker-compose file it became apparent that we needed to specify a version of mysql `5.7`.  The reason why may be due to [one of these reasons](https://github.com/trilogy-libraries/trilogy/issues?q=is%3Aissue+is%3Aopen+mysql+8+). I didn't spend too much time digging in.

~~I also spend some time to update the array to all be floats, because having a string in the array didn't feel quite right 😄~~ https://github.com/composite-primary-keys/composite_primary_keys/pull/589. I also updated the actions checkout to the latest version.

### DIFF
--- a/.github/workflows/mariadb.yml
+++ b/.github/workflows/mariadb.yml
@@ -8,7 +8,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        ruby: [2.7, "3.0", 3.1, 3.2]
+        ruby: [2.7, 3.0, 3.1, 3.2]
     runs-on: ${{matrix.os}}
     services:
       mariadb:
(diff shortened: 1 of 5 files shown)


.github/workflows/mariadb.yml
```diff
@@ -8,7 +8,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        ruby: [2.7, "3.0", 3.1, 3.2]
+        ruby: [2.7, 3.0, 3.1, 3.2]
```
cfis, line 11: See https://github.com/composite-primary-keys/composite_primary_keys/pull/589

zmariscal: @cfis thanks for the PR reference. Updated :)

cfis:
Thanks for the update!

Issue #474, opened 2019-02-15
Rails 6 support


diguliu:
Is there any work being done towards Rails 6 support? Also, is there a workaround to use it with Rails 6 beta right now?

javierjulio:
@diguliu #472

PR #613, opened 2023-12-02
Fix Error for Array-Based Composite Keys in nested_attribute

yebityon:
# Problem Description

I recently updated my project to Rails 6 and faced a testing error when passing a composite key as an array to nested_attribute.

I have added a test case for this error. Please refer to this [test case](https://github.com/composite-primary-keys/composite_primary_keys/commit/41f81c8044776bc8d9c69b071386050e8dc391c1) if necessary.

# PR Description

Since [this commit](https://github.com/composite-primary-keys/composite_primary_keys/issues/474), record detection in nested_attributes.rb has been conducted using the `detect` method instead of `cpk_detect_record`.

I could not find any mention of this change, and considering that `cpk_detect_record` still exists in `nested_attributes.rb` but is not used, I created this PR.

This PR reverts to the approach used prior to the introduction of this [PR](https://github.com/composite-primary-keys/composite_primary_keys/commit/846b5e86089f866556769658de53ff9d9f82ff52#diff-f546fbccae6c9d276ea04e316415d39c2cc08c040f23119978d57f1fe8197406).

If we are intended to use `CompositePrimaryKeys::CompositeKeys` for creating primary keys as demonstrated in the test case (i.e., if the above change was intentional), I will close this PR.

Also, a quick thanks to the maintainers for their ongoing work on the project.

### DIFF
--- a/lib/composite_primary_keys/nested_attributes.rb
+++ b/lib/composite_primary_keys/nested_attributes.rb
@@ -66,7 +66,7 @@ module ActiveRecord
             # Make sure we are operating on the actual object which is in the association's
             # proxy_target array (either by finding it, or adding it if not found)
             # Take into account that the proxy_target may have changed due to callbacks
-             target_record = cpk_detect_record(attributes['id'], association.target)
+            target_record = cpk_detect_record(attributes['id'], association.target)
             if target_record
               existing_record = target_record
             else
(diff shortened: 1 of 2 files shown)


cfis:
Looks good to me - thanks!

PR #615, opened 2024-08-28
Fix scope overriding join relation query

henriquetatagiba:
Adding scopes after the join relation removes the join query from the generated SQL.

Rearrange code to match [rails 7.0 method](https://github.com/rails/rails/blob/7-0-stable/activerecord/lib/active_record/reflection.rb#L175)

Example:
```rb
class User < ApplicationRecord
  has_many :features, -> { enabled }
end

class Feature < ApplicationRecord
  scope :enabled, -> { where.not(user_id: nil).where(enabled: true) }
end

# User.joins(:features).to_sql
# INNER JOIN query won't include `users.id = features.user_id` but only `features.user_id IS NOT NULL`
# when the expected is both
```

### DIFF
--- a/lib/composite_primary_keys/reflection.rb
+++ b/lib/composite_primary_keys/reflection.rb
@@ -6,6 +6,12 @@ module ActiveRecord
         scope_chain_items = join_scopes(table, predicate_builder)
         klass_scope       = klass_join_scope(table, predicate_builder)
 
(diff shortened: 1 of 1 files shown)


cfis:
Sorry for not responding to this earlier. Thanks for the fix!

Issue #616, opened 2024-12-13
CPK and hash-partitioned table in PostgreSQL - building records via parent association doesn't work


glizda:
## Issue reproduced in the Ruby script that you can find at the and of my message

The ruby script at the end of my message (based on [active_record bug report template](https://github.com/rails/rails/blob/main/guides/bug_report_templates/active_record.rb)) presents a simplified version of my setup together with test cases that show my issue.

## Issue description

I have two tables - `bookings` and `booking_time_segments`.

`booking_time_segments` is partitioned using hash strategy (partition key `booking_id`) and has composite primary key (`booking_id, id`).

Here are the AR models:
```ruby
class Booking < ActiveRecord::Base
  has_many :booking_time_segments, autosave: true
end

class BookingTimeSegment < ActiveRecord::Base
  self.primary_keys = :booking_id, :id
  belongs_to :booking
end
```

In that setup, I do:
```ruby
booking = Booking.new

bts_params = [
  { duration_in_seconds: 100, sorting_id: 0 },
  { duration_in_seconds: 200, sorting_id: 1 },
  { duration_in_seconds: 300, sorting_id: 2 }
]

booking.booking_time_segments.build(bts_params)
```

The line `booking.booking_time_segments.build(bts_params)` returns (as expected) the array of three `BookingTimeSegment` objects, suggesting that they are being properly built based on the params:
```ruby
[#<BookingTimeSegment id: nil, duration_in_seconds: 100, sorting_id: 0, booking_id: nil, created_at: nil, updated_at: nil>, #<BookingTimeSegment id: nil, duration_in_seconds: 200, sorting_id: 1, booking_id: nil, created_at: nil, updated_at: nil>, #<BookingTimeSegment id: nil, duration_in_seconds: 300, sorting_id: 2, booking_id: nil, created_at: nil, updated_at: nil>]
```

However, they don't really get built and/or assigned properly. If I then call `booking.booking_time_segments` (on not yet persisted `booking`), I get only one child record (the last one from the previous array):
```ruby
#<ActiveRecord::Associations::CollectionProxy [#<BookingTimeSegment id: nil, duration_in_seconds: 300, sorting_id: 2, booking_id: nil, created_at: nil, updated_at: nil>]>
```

Later saving the `booking` record only creates one associated `booking_time_segment` record. My expectation is that all three `booking_time_segments` records get properly built, assigned to the `booking` parent and persisted when `booking` is persited.

## Question

[...]

PR #617, opened 2024-12-16
AR 6.1: Fix object comparison for models with composite keys

glizda:
Addresses issue https://github.com/composite-primary-keys/composite_primary_keys/issues/616

## Overview - overwriting `==`

In `ActiveRecord` two new (not yet persisted) AR objects are _not_ considered equal, even if all their attributes (including primary key) are `nil`.

We want the same behaviour for models with composite primary key handled by `composite_primary_keys` gem. To do that, we override `==` method. In the [AR source](https://github.com/rails/rails/blob/v6.1.7.10/activerecord/lib/active_record/core.rb#L637-L643), there is a special key for primary key being empty (check using `!id.nil?`). For composite primary key, we will use `id.all? {|key| !key.nil?}` to check that.

I also took inspiration from [AR 7.2](https://github.com/rails/rails/blob/33beb0a38db1c058123a8e3cc298cad918adfe32/activerecord/lib/active_record/attribute_methods/composite_primary_key.rb#L16-L22) where a helper method `primary_key_values_present?` is defined that differs based on whether the primary key is singular or composite.

## Overwriting `hash`

`hash` method on AR model also uses `!id.nil?` and thus also doesn't work well for models with composite primary key. it also gets improved in this PR.

## Tests

I wrote tests both for models with single and with composite primary keys. I switched to using `Restaurant` model instead of `Capitol` because I wanted a table that has a composite primary key _and_ some other columns (for the purposes of `test_composite_persisted_object_hash_is_based_on_primary_key` test case).

### DIFF
--- a/lib/composite_primary_keys/core.rb
+++ b/lib/composite_primary_keys/core.rb
@@ -16,6 +16,29 @@ module ActiveRecord
       super
     end
 
+    def primary_key_values_present?
+      if self.composite?
+        id.all? {|key| !key.nil?}
+      else
+        !id.nil?
+      end
+    end
+
+    def ==(comparison_object)
+      super ||
+        comparison_object.instance_of?(self.class) &&
+        primary_key_values_present? &&
+        comparison_object.id == id
+    end
+
+    def hash
+      if primary_key_values_present?
+        self.class.hash ^ id.hash
+      else
+        super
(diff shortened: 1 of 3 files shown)


cfis:
Thanks!

Issue #619, opened 2024-12-18
ActiveRecord 6.1 - `update_all` doesn't update loaded association records


glizda:
## Overview

In ActiveRecord 6.1, calling `update_all` on association (ex. `user.readings.update_all(rating: 3)`) not only updates the records in the DB, but also updates the loaded association objects. In other words, when you then use `user.readings`, they will have updated `rating` values.

When `composite_primary_key` gem is used, this behaviour is different, _even for models without composite primary key defined_ (because `update_all` in `ActiveRecord::Relation` is overwritten for all AR models).

## Example

Buy commenting out `gem "composite_primary_keys", '~> 13.0.0'` you can get the second test to pass, showing that the `update_all` behaviour is different with and without `composite_primary_key` event for models with single primary key.

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  ruby "2.7.7"

  gem "composite_primary_keys", '~> 13.0.0'
  gem "pg", "~> 1.5.3"
  gem "rails", "~> 6.1.7.4"
end

require "active_record/railtie"
require "minitest/autorun"

class TestApp < Rails::Application
  config.load_defaults Rails::VERSION::STRING.to_f
  config.eager_load = false
  config.logger = Logger.new($stdout)

  ActiveRecord::Base.establish_connection(
    adapter: "postgresql",
    database: "composite_pk_test",
    host: "localhost",
    username: "postgres"
  )
end

ActiveRecord::Schema.define do
  drop_table :users, if_exists: true
  drop_table :readings, if_exists: true

  execute <<-SQL
    create table users (
      id serial not null,
      primary key (id)
   );

    create table readings (
      id serial not null,
      user_id int not null,
      rating int not null,
      primary key (id)
    );
  SQL
end

class User < ActiveRecord::Base
  has_many :readings
end

class Reading < ActiveRecord::Base
  belongs_to :user
end

class BugTest < ActiveSupport::TestCase
  def test_update_all_updates_db_records
    user = User.create
    user.readings.create(rating: 4)
    user.readings.create(rating: 5)

    assert_equal [4, 5], user.readings.map(&:rating)

    user.readings.update_all(rating: 3)

    # Reload to check that the records were updated in the DB
    user.readings.reload
    assert_equal [3, 3], user.readings.map(&:rating)
  end

  def test_update_all_updates_loaded_association
    user = User.create
    user.readings.create(rating: 4)
[...]

PR #620, opened 2024-12-18
AR 6.1: Fix `update_all` to update loaded associations

glizda:
Addresses https://github.com/composite-primary-keys/composite_primary_keys/issues/619

## Overview

In ActiveRecord 6.1, calling `update_all` on association (ex. `user.readings.update_all(rating: 3)`) not only updates the records in the DB, but also updates the loaded association objects. In other words, when you then use `user.readings`, they will have updated `rating` values.

When `composite_primary_key` gem is used, this behaviour is different, _even for models without composite primary key defined_ (because `update_all` in `ActiveRecord::Relation` is overwritten for all AR models).

This PR fixes it. I changed the `update_all` and `delete_all` methods to more closely resemble those from [Rails 6.1 source](https://github.com/rails/rails/blob/6-1-stable/activerecord/lib/active_record/relation.rb) (with the important difference of how `where` constraints are built for composite primary key models).

### DIFF
--- a/lib/composite_primary_keys/relation.rb
+++ b/lib/composite_primary_keys/relation.rb
@@ -19,23 +19,22 @@ module ActiveRecord
     def update_all(updates)
       raise ArgumentError, "Empty list of attributes to change" if updates.blank?
 
-      if eager_loading?
-        relation = apply_join_dependency
-        return relation.update_all(updates)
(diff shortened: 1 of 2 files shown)


cfis:
Cool - thanks!

PR #621, opened 2024-12-30
Ignore orders for delete_all/update_all queries.

Jack12816:
**- What is it good for**

While upgrading an application from Rails 5.2 to 6.1 (CPK 11.2 -> 13.0) some specs broke. I tracked it down to `Entity.delete_all` calls on the setup of the test suite. Then I found the broken SQL statement and about an syntax error near to `ORDER` - there is nothing to order at the top-level for `DELETE` and `UPDATE` queries. This is caused by a `default_scope { order(created_at: :asc) }`.

<details>
  <summary>Click for Details</summary>

### ActiveRecord::Base.delete_all

```ruby
class Tariff < ActiveRecord::Base
  default_scope { order(created_at: :asc) }
end

Tariff.delete_all
```

```sql
DELETE FROM "tariffs"
WHERE ("tariffs"."tariff_id", "tariffs"."start_date") IN (
  SELECT "tariffs"."tariff_id", "tariffs"."start_date"
  FROM "tariffs"
  ORDER BY "tariffs"."created_at" ASC
)
ORDER BY "tariffs"."created_at" ASC -- syntax error due to the
                                    -- default scope/order

-- ActiveRecord::StatementInvalid: PG::SyntaxError:
--   ERROR:  syntax error at or near "ORDER"
-- LINE 1: ...OM "tariffs" ORDER BY "tariffs"."created_at" ASC) ORDER BY "...
```

### ActiveRecord::Base.update_all

```ruby
class Tariff < ActiveRecord::Base
  default_scope { order(created_at: :asc) }
end

Tariff.update_all(amount: 1337)
```

```sql
UPDATE "tariffs"
WHERE ("tariffs"."tariff_id", "tariffs"."start_date") IN (
  SELECT "tariffs"."tariff_id", "tariffs"."start_date"
  FROM "tariffs"
  ORDER BY "tariffs"."created_at" ASC
)
ORDER BY "tariffs"."created_at" ASC -- syntax error due to the
                                    -- default scope/order

-- ActiveRecord::StatementInvalid: PG::SyntaxError:
--   ERROR:  syntax error at or near "ORDER"
-- LINE 1: ...OM "tariffs" ORDER BY "tariffs"."created_at" ASC) ORDER BY "...
```

</details>

**- What I did**

Just removed the ordering on `delete_all` and `update_all` SQL statement assembly and added tests for it. I fixed the issue at the `ar_6.1.x` branch, but I'm sure this must be ported back to 6.0 and  master. 

**- A picture of a cute animal (not mandatory but encouraged)**
image

### DIFF
--- a/lib/composite_primary_keys/relation.rb
+++ b/lib/composite_primary_keys/relation.rb
@@ -37,7 +37,6 @@ module ActiveRecord
 
       stmt.take(arel.limit)
       stmt.offset(arel.offset)
-      stmt.order(*arel.orders)
 
       if updates.is_a?(Hash)
         if klass.locking_enabled? &&
@@ -81,7 +80,6 @@ module ActiveRecord
 
       stmt.take(arel.limit)
       stmt.offset(arel.offset)
-      stmt.order(*arel.orders)
 
       affected = klass.connection.delete(stmt, "#{klass} Destroy")
 
--- a/test/fixtures/tariff.rb
+++ b/test/fixtures/tariff.rb
@@ -1,5 +1,6 @@
 class Tariff < ActiveRecord::Base
-	self.primary_keys = [:tariff_id, :start_date]
-	has_many :product_tariffs, :foreign_key => [:tariff_id, :tariff_start_date], :dependent => :delete_all
-	has_many :products, :through => :product_tariffs, :foreign_key => [:tariff_id, :tariff_start_date]
+  self.primary_keys = [:tariff_id, :start_date]
+  has_many :product_tariffs, :foreign_key => [:tariff_id, :tariff_start_date], :dependent => :delete_all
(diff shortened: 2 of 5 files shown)


cfis:
The calls to `stmt.order(*arel.orders)`  are copied from ActiveRecord. See the original method here:

https://github.com/rails/rails/blob/6-1-stable/activerecord/lib/active_record/relation.rb#L455

Removing the order call would likely will break any non-CPK AR classes. So we can't do that. I also imagine the `stmt.offset` and `stmt.take` calls cause the same issue type of issue.

The PR needs to be updated to not break non-CPK models.

Jack12816:
Closing for the same reason: https://github.com/composite-primary-keys/composite_primary_keys/pull/622#issuecomment-2574395943

PR #624, opened 2025-12-11
Quote Column Names

Hamms:
The current MySQL query generation logic does not quote the primary keys in generated subqueries, which can result in queries like:

```mysql
DELETE FROM `datablock_storage_kvps` WHERE (datablock_storage_kvps.project_id, datablock_storage_kvps.key) IN (SELECT project_id,key FROM (SELECT DISTINCT `datablock_storage_kvps`.`project_id`, `datablock_storage_kvps`.`key` FROM `datablock_storage_kvps` WHERE `datablock_storage_kvps`.`project_id` = 13) __active_record_temp)
```

Which can, depending on the names given to the keys (in this case, "key"), result in errors like:

    ActiveRecord::StatementInvalid: Mysql2::Error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'key FROM (SELECT DISTINCT `datablock_storage_kvps`.`project_id`, `datablock_stor' at line 1

This PR adds logic to quote the parameters in both the `WHERE` and `SELECT` clauses, resulting in a query like:

```mysql
DELETE FROM `datablock_storage_kvps` WHERE (`datablock_storage_kvps`.`project_id`, `datablock_storage_kvps`.`key`) IN (SELECT `project_id`,`key` FROM (SELECT DISTINCT `datablock_storage_kvps`.`project_id`, `datablock_storage_kvps`.`key` FROM `datablock_storage_kvps` WHERE `datablock_storage_kvps`.`project_id` = 13) __active_record_temp)
```

### DIFF
--- a/lib/composite_primary_keys/arel/to_sql.rb
+++ b/lib/composite_primary_keys/arel/to_sql.rb
@@ -5,9 +5,9 @@ module Arel
         values = o.map do |key|
           case key
             when Arel::Attributes::Attribute
-              "#{key.relation.name}.#{key.name}"
+              "#{quote_table_name(key.relation.name)}.#{quote_column_name(key.name)}"
             else
-              key
+              quote_column_name(key)
           end
         end
         collector << "(#{values.join(', ')})"
--- a/lib/composite_primary_keys/relation.rb
+++ b/lib/composite_primary_keys/relation.rb
@@ -173,8 +173,8 @@ module ActiveRecord
     #
     # UPDATE `reference_codes`
     # SET ...
-    # WHERE (reference_codes.reference_type_id, reference_codes.reference_code) IN
-    #  (SELECT reference_type_id,reference_code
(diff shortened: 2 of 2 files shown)


cfis:
Ok, well tests pass, so good by me. Let me know once you take this out of draft status.

cfis:
A test is always good!

Hamms:
Would you say a test for this is worth adding yet another text fixture? I made a couple of attempts to create a test case which reproduces the original error by modifying existing fixtures, but kept breaking other tests in the process :upside_down_face:

I can easily create a new fixture just for this test, but I don't want to add unnecessarily to the maintenance burden here. What's your preference?

cfis:
Good question - I was trying to avoid adding new fixtures. But now that CPK has been superseded by what is built into ActiveRecord, I leave it up to you.

